PR #14, opened 2023-10-06
feat: support eslint-define-config rules

Shinigami92:
See https://github.com/Shinigami92/eslint-define-config/pull/238 for more details

- https://github.com/Shinigami92/eslint-define-config/pull/238


packages/eslint-plugin-ts/src/eslint-define-config-support.d.ts
```diff
@@ -0,0 +1,473 @@
+declare module 'eslint-define-config' {
```
antfu, line 1: I wonder if this would be a `types.d.ts` file under each function's directory and then we could bundle them into a single file like this. For better domain separation.

Also, do you have an idea in mind to support plugin renaming with the Flat Config?

We have a union package that merges js and ts rules into a single one: https://eslint.style/packages/default, meaning the prefix would be `@stylistic/` instead of `@stylistic/ts/` for them. Users could also rename to arbitrary prefixes, for example in https://github.com/antfu/eslint-config#plugins-renaming I renamed `@typescript-eslint/*` to `ts/*`.

Shinigami92: > I wonder if this would be a `types.d.ts` file under each function's directory and then we could bundle them into a single file like this. For better domain separation.

So is this the same idea as I already wrote as TODO here?

https://github.com/Shinigami92/eslint-stylistic/blob/67337ad6f44a66882280962cd4d9382d35e08072/scripts/migrate.ts#L296-L297

Just to be clear, I could write either `packages/eslint-plugin-ts/rules/block-spacing/types.d.ts` or I could write `packages/eslint-plugin-ts/src/rules/block-spacing/types.d.ts`
What was the reason for generating files outside the `src` folder?

> Also, do you have an idea in mind to support plugin renaming with the Flat Config?
> 
> We have a union package that merges js and ts rules into a single one: [eslint.style/packages/default](https://eslint.style/packages/default), meaning the prefix would be `@stylistic/` instead of `@stylistic/ts/` for them. Users could also rename to arbitrary prefixes, for example in [antfu/eslint-config#plugins-renaming](https://github.com/antfu/eslint-config#plugins-renaming) I renamed `@typescript-eslint/*` to `ts/*`.

I have 0 knowledge yet about `FlatConfig`, aside that it is in an _experimental not ready for support yet_ state :thinking: 
The whole `FlatConfig` support in `eslint-define-config` is 100% driven by @sxzz

TBH, I would like to first make the support ready for the current config format, because this is what is used by (I feel like) 99% of eslint's userbase.
For renaming the rules, I would suggest something like https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#key-remapping-via-as when it is really needed. But for now I suggest using one way, because otherwise you would also start to confuse the userbase about what is the "correct" way now to configure "that one rule".

antfu: > packages/eslint-plugin-ts/rules/block-spacing/types.d.ts

Yes, I think that sounds good.

> What was the reason for generating files outside the src folder?

I just want to keep the structure simple, as we also grab files from `rules/` for documentation. The codes are generated/migrated only because the list is final yet, we could start to maintain them by hand once ESLint deprecated them. In that way we don't introduce misalignment before that happens.

> Flat Config

Flat Config is basically the same as the current one, only that the `plugins` field is an object instead of an array:

```ts
// old
defineConfig({
  plugins: [
    'vue',
  ]
})

// new
import vuePlugin from 'eslint-plugin-vue'

defineConfig([{
  plugins: {
    vue: vuePlugin
  }
}])
```

So in Flat Config, you define the name of each plugin, instead of strictly following the package name.

Going even further, I would imagine that `defineConfig` could be agnostic and automatically infer from the plugin object.

Just an idea here, ofc not blocking as we could do type mapping as an workaround for now.

Shinigami92: Okay, understand

So right now https://github.com/Shinigami92/eslint-define-config/pull/238 is just the start for supporting augmented global properties for rules

So the plugins property of the config is currently not affected by this at all

However, YES you are right in terms of in the long run I want to support customizing much more stuff like e.g. `settings`, `plugins`, `extends`, `parserOptions` and `parser`
But I will work on this when `rules` are done and working correctly :slightly_smiling_face: 

_please remember, I'm not a machine like you senpai_ :rofl:

Shinigami92:
Currently we would need to import the js support into the index, but it's written in .js instead of .ts :thinking: 
So I would love if we rewrite the js stuff into ts at all

Also, right now I did not generate the JSDoc stuff

coderwyd (commented):
图片

图片


Incorrectly referenced path.

Shinigami92:
> 图片
> 
> 图片
> 
> Incorrectly referenced path.

We discussed anyway that we want to potentially not export it via index but by `<plugin>/define-config-support` so you wont bleed every types into your code, but only the ones that are relevant
See https://github.com/eslint-types/eslint-define-config/pull/238/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5 for more details

coderwyd:
> > 图片
> > 图片
> > Incorrectly referenced path.
> 
> We discussed anyway that we want to potentially not export it via index but by `<plugin>/define-config-support` so you wont bleed every types into your code, but only the ones that are relevant See https://github.com/eslint-types/eslint-define-config/pull/238/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5 for more details

Understood, so should we delete this line?

Shinigami92:
> > > 图片
> > > 图片
> > > Incorrectly referenced path.
> > 
> > 
> > We discussed anyway that we want to potentially not export it via index but by `<plugin>/define-config-support` so you wont bleed every types into your code, but only the ones that are relevant See [eslint-types/eslint-define-config#238 (files)](https://github.com/eslint-types/eslint-define-config/pull/238/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) for more details
> 
> Understood, so should we delete this line?

yes + rename the file and export it from the `exports` in `package.json`

Issue #15, opened 2023-10-07
Docs style is confusing


chizukicn:
**Reproduction**
When accessing any path that does not exist under the document path, the 404 page will be matched to the home page, and style confusion will occur when entering other pages.

**Environments**

macOS 14 , Microsoft Edge latest version and Chrome latest version

![c8efbef975472c58785b7da7edc69ed0](https://github.com/eslint-stylistic/eslint-stylistic/assets/16945858/67bbdad1-976b-4c1c-831f-e0676026741f)

PR #16, opened 2023-10-07
docs: fix 404 page cant match

chizukicn:
fix https://github.com/eslint-stylistic/eslint-stylistic/issues/15


netlify.toml
```diff
@@ -1,12 +1,9 @@
 [build.environment]
   NODE_VERSION = "18"
+  NODE_OPTIONS = "--max_old_space_size=4096"
```
kecrily, line 3: Why would we need to add this? I mean why would this cause a JavaScript heap out of memory?

netlify.toml
```diff
@@ -1,12 +1,9 @@
...
 
 [build]
   publish = "docs/.vitepress/dist"
   command = "pnpm run docs:build"
 
-[[redirects]]
-  from = "/*"
-  to = "/index.html"
-  status = 200
```
antfu, line 11: I would prefer VitePress to fix this instead of workaround here.

netlify.toml
chizukicn: https://app.netlify.com/sites/eslint-stylistic/deploys/6521011f2f69ed0008616702
https://github.com/unocss/unocss/commit/b07065aa1f994c8f41ea888582abf2aa405ae3df
This problem has also occurred in unocss document packaging. I am not sure if it is a problem with vitepress.

netlify.toml
userquin: @antfu VP cannot fix this since Netlify is responding with 200 forwarding index.html content (the missing page is there), you should remove SPA from Netlify, then Netlify will send 404 and VP will show 404 page. 

If you mean VP detects that the route doesn't exist and redirects to 404, I'll take a look at that later.

FYI: I've removed SPA from Vite PWA and Vitest docs (and also from my vuetify nuxt module).

antfu: What do you mean by "remove SPA"? Removing this redirection would just work?

userquin: yes, check this url https://deploy-preview-16--eslint-stylistic.netlify.app/guide/missing and then navigate, check the same url in production

PR #20, opened 2023-10-11
feat: define-config-support files

Shinigami92:
see discussion in https://github.com/eslint-stylistic/eslint-stylistic/pull/14#issuecomment-1754484046


antfu:
Why move from ts to js package? Tests seems to fail

Shinigami92:
> Why move from ts to js package? Tests seems to fail

It's not a move
It's also now for js

Shinigami92:
I don't know why the tests are failing, the CI dont show either what's happening 
I only committed files that are relevant 
I assume it has something to do with external forks + PRs
These CI steps are only executed on PRs
Could you have a look yourself?

PR #22, opened 2023-10-16
feat(eslint-plugin-jsx): add jsx-self-closing-comp

zhiyuanzmj:
# Disallow extra closing tags for components without children (`react/self-closing-comp`)

🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).

Components without children can be self-closed to avoid unnecessary extra closing tag.
## Rule Details

Examples of **incorrect** code for this rule:

```jsx
var HelloJohn = <Hello name="John"></Hello>;

var HelloJohnCompound = <Hello.Compound name="John"></Hello.Compound>;
```

Examples of **correct** code for this rule:

```jsx
var contentContainer = <div className="content"></div>;

var intentionalSpace = <div>{' '}</div>;

var HelloJohn = <Hello name="John" />;

var HelloJohnCompound = <Hello.Compound name="John" />;

var Profile = <Hello name="John"><img src="picture.png" /></Hello>;

var ProfileCompound = <Hello.Compound name="John"><img src="picture.png" /></Hello.Compound>;

var HelloSpace = <Hello>{' '}</Hello>;
```

## Reference
https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/self-closing-comp.md


antfu:
Looks good. I think we'd better to rename the rule to `jsx-self-closing-comp` to have a clear scope.

PR #40, opened 2023-11-04
refactor: rename func-call-spacing to function-call-spacing

### LINKED ISSUE #39: Rename func-call-spacing to function-call-spacing
## Overview

Moving from Discord, I propose renaming of rule `func-call-spacing` to `function-call-spacing` for the following reasons:

1. Inconsistent naming scheme based on the following function-related rule name:
  a. `function-call-argument-newline`
  b. `function-paren-newline`
2. Clearer and more concise rule name. This rule is referring to functions and not to be confused with similar-sounding terms like func*tor*

Namchee:
## Overview

Closes #39 

This pull request renames `func-call-spacing` function to `function-call-spacing` for the reasons stated in #39. Generally, it only renames all the files and references of the said rules.

Do note that this change won't be applied on migration since it doesn't touch the migration scripts at all.


ota-meshi:
I think removing the original rule is a breaking change. I think it's better to leave the original rule as an alias. What do you think?

Namchee:
@ota-meshi, valid point. For now, I'll duplicate the plugin under a new name.

Namchee:
Dropped in favor of #63

PR #49, opened 2023-11-07
perf: no longer rely on `jsx-ast-utils`, reduce install size

### LINKED ISSUE #43: Reduce amount of unneeded third-party polyfills
`@stylistic/eslint-plugin-jsx` depends on `jsx-ast-util` which brings a lot of useless polyfills like `is-string`, `object-keys`, `object.values`, `array-includes` and many more. In total I estimatea at least 30-40 packages useless packages that no modern JS engine needs and which slow down the installation time of this plugin.

antfu:
It turns out that we are not using a lot of it, we could completely remove it to reduce the install size. Close #43

### DIFF
--- a/packages/eslint-plugin-jsx/package.json
+++ b/packages/eslint-plugin-jsx/package.json
@@ -39,8 +39,7 @@
   },
   "dependencies": {
     "@stylistic/eslint-plugin-js": "workspace:^",
-    "estraverse": "^5.3.0",
-    "jsx-ast-utils": "^3.3.5"
+    "estraverse": "^5.3.0"
   },
   "devDependencies": {
     "@babel/eslint-parser": "^7.22.15",
--- a/packages/eslint-plugin-jsx/rollup.config.mts
+++ b/packages/eslint-plugin-jsx/rollup.config.mts
@@ -29,8 +29,6 @@ export default defineConfig({
   external: [
     ...Object.keys(pkg.dependencies || []),
     ...Object.keys(pkg.peerDependencies || []),
-    'jsx-ast-utils/propName',
-    'jsx-ast-utils/elementType',
     'node:process',
   ],
 })
--- a/packages/eslint-plugin-jsx/rules/jsx-sort-props/jsx-sort-props.js
+++ b/packages/eslint-plugin-jsx/rules/jsx-sort-props/jsx-sort-props.js
@@ -5,8 +5,6 @@
 
 'use strict'
 
-const propName = require('jsx-ast-utils/propName')
-
 const includes = (arr, value) => arr.includes(value)
 const toSorted = (arr, compareFn) => [...arr].sort(compareFn)
 
@@ -58,8 +56,8 @@ function shouldSortToEnd(node) {
 }
 
 function contextCompare(a, b, options) {
-  let aProp = propName(a)
-  let bProp = propName(b)
+  let aProp = jsxUtil.getPropName(a)
+  let bProp = jsxUtil.getPropName(b)
 
   const aSortToEnd = shouldSortToEnd(a)
   const bSortToEnd = shouldSortToEnd(b)
@@ -416,8 +414,8 @@ module.exports = {
           if (decl.type === 'JSXSpreadAttribute')
             return attrs[idx + 1]
 
-          let previousPropName = propName(memo)
-          let currentPropName = propName(decl)
+          let previousPropName = jsxUtil.getPropName(memo)
+          let currentPropName = jsxUtil.getPropName(decl)
           const previousValue = memo.value
           const currentValue = decl.value
           const previousIsCallback = isCallbackPropName(previousPropName)
--- a/packages/eslint-plugin-jsx/util/jsx.js
+++ b/packages/eslint-plugin-jsx/util/jsx.js
@@ -4,8 +4,6 @@
 
 'use strict'
 
-const elementType = require('jsx-ast-utils/elementType')
-
 const astUtil = require('./ast')
 const isCreateElement = require('./isCreateElement')
 const variableUtil = require('./variable')
@@ -20,7 +18,7 @@ const COMPAT_TAG_REGEX = /^[a-z]/
  * @returns {boolean} Whether or not the node corresponds to a DOM element.
  */
 function isDOMComponent(node) {
-  const name = elementType(node)
+  const name = getElementType(node)
   return COMPAT_TAG_REGEX.test(name)
 }
 
@@ -183,6 +181,57 @@ function isReturningOnlyNull(ASTnode, context) {
   return found && !foundSomethingElse
 }
 
+/**
+ * Ported from `jsx-ast-utils/propName` to reduce bundle size
+ */
+function getPropName(...args) {
+  const prop = args.length > 0 && args[0] !== undefined ? args[0] : {}
+  if (!prop.type || prop.type !== 'JSXAttribute')
+    throw new Error('The prop must be a JSXAttribute collected by the AST parser.')
+  if (prop.name.type === 'JSXNamespacedName')
+    return `${prop.name.namespace.name}:${prop.name.name.name}`
+  return prop.name.name
+}
+
+function resolveMemberExpressions(...args) {
+  const object = args.length > 0 && args[0] !== undefined ? args[0] : {}
+  const property = args.length > 1 && args[1] !== undefined ? args[1] : {}
+
+  if (object.type === 'JSXMemberExpression')
+    return `${resolveMemberExpressions(object.object, object.property)}.${property.name}`
+
+  return `${object.name}.${property.name}`
+}
+
+/**
+ * Returns the tagName associated with a JSXElement.
+ * Ported from `jsx-ast-utils/elementType` to reduce bundle size
+ */
+function getElementType(...args) {
+  const node = args.length > 0 && args[0] !== undefined ? args[0] : {}
+  const name = node.name
+
+  if (node.type === 'JSXOpeningFragment')
+    return '<>'
+
+  if (!name)
+    throw new Error('The argument provided is not a JSXElement node.')
+
+  if (name.type === 'JSXMemberExpression') {
+    const _name$object = name.object
+    const object = _name$object === undefined ? {} : _name$object
+    const _name$property = name.property
+    const property = _name$property === undefined ? {} : _name$property
+
+    return resolveMemberExpressions(object, property)
+  }
+
+  if (name.type === 'JSXNamespacedName')
+    return `${name.namespace.name}:${name.name.name}`
+
+  return node.name.name
+}
+
 module.exports = {
   isDOMComponent,
   isFragment,
@@ -191,4 +240,5 @@ module.exports = {
   isWhiteSpaces,
   isReturningJSX,
   isReturningOnlyNull,
+  getPropName,
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -185,9 +185,6 @@ importers:
       estraverse:
         specifier: ^5.3.0
         version: 5.3.0
-      jsx-ast-utils:
-        specifier: ^3.3.5
-        version: 3.3.5
     devDependencies:
       '@babel/eslint-parser':
         specifier: ^7.22.15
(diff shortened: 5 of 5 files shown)


ljharb:
I think it is a very shortsighted decision to inline a dep to worry about something as inconsequential as "install size", especially for a dev tool where disk space is largely infinite and free.

antfu:
While I generally agree your arguments about the "correctness" in #43, I can't agree your statement about "disk space is largely infinite and free."

Correctness is indeed important - we have tests to ensure that. But blindly don't care about both "install size" and "runtime cost" due to "correctness" is honestly a bit extreme. I would personally try to approach the best on both side - the code can be correct, efficient and small when giving enough care.

In this specific case, we are only using `propName` and `elementType` functions - which are simple and pure. The scope is pretty clear and I don't see any "shortsighted" fault it could introduce. Asking **everybody** to download extra 60MB for the code that is never used is much more irresponsible to me.

ljharb:
I would be happy to extract each of those to a separate package, rather than inlining the code here - would that help ensure implementations can be shared instead of vendored?

antfu:
In total they are only 50 LoC, pure and well-scoped. Why do you think it's better to serve as separate packages? If so, what would be the difference from the already-inlined utils like `isCreateElement` or `isFirstLetterCapitalized`?

ljharb:
That way if there's any fixes, we can both benefit from them without having to remember to update the vendored copies here - same reason vendoring code is a bad practice.

If those already-inlined utils can be shared between jsx-ast-utils and this project, i'd be happy to extract those as well. The entire ecosystem benefits when well-tested code is maximally shared.

antfu:
My concerns are:

- Packages have overhead, both on installing and runtime resolving
- Granularity is hard to balance for such utils. Either we go all-in-one like `lodash`, or we go one-function-package like `is-string`. The fromer is inefficient in downloading and init unrelated code, the latter is inefficient in having too much package overhead.
- CJS/ESM interop can be a problem. For example, `require('jsx-ast-utils/elementType')` seems to be a scoped sub-entry import, but it's actually a one-line re-export from the entire package. In bundling, it bails out tree-shaking, and in runtime, it forces Node.js to initialize the entire package and its deps, despite we only need a pure function.

I am thinking of rewriting the whole project to use ESM and TS, step by step. And remove unused inline utils gradually (as we only care about formatting but many utils are checking the semantics).

If the shared packages can provide clean ESM for us to consume, I am happy to have it and collaborate. But even so, I guess the list of shared functions needs to wait until we do a few rounds of reactors and clean-ups to have a clear image of what we really need. In that regard, I think it's better to have them inline to iterate faster and extract them when we are settled.

ljharb:
In my experience package overhead is exceedingly minimal compared to all other concerns.

I'm not sure why the format of the package is relevant as long as it can be `import`ed, which CJS can be.

Inlining as an interim step makes sense, but if we're in an interim step, why does install size or perf matter at all? That can be focused on once the functionality is complete.

antfu:
The format matters because if `jsx-ast-utils` provides tree-shakable ESM, today I don't need to vendor it - I can just bundle what I need (those 50 LoC), keep the install size small while having a single source code base to maintain and share.

We are not in an interim step, we shipped 1.0 and ppl started migrating. We have people creating #43, which is a direct indicator that the install size matters. We want the first version to be 1:1 equivalent to make the migration easier, but on the other hand, we are just peeled from huge repos that we have many redundant and inefficient parts that we need to refactor. Many files/functions under `packages/eslint-plugin-jsx/util` are unused but it takes manual efforts to clean them and reduce the maintenance scope for the future.

ljharb:
CJS is tree-shakeable, and if you deep import the files you need from the package (regardless of format) then tree-shaking can't beat it (and can't achieve that anyways).

The install size would be largely addressed by using extracted packages.

SukkaW:
> The format matters because if `jsx-ast-utils` provides tree-shakable ESM, today I don't need to vendor it - I can just bundle what I need (those 50 LoC), keep the install size small while having a single source code base to maintain and share.

@antfu IMHO it is better to stick with the `jsx-ast-utils`, in case any future behavior changes.

If we bundle `jsx-ast-utils` directly, the final output size would be 72.6 KiB: [bundlejs.com](https://bundlejs.com/?q=jsx-ast-utils).

But if we alias all polyfills with `nolyfill` during the bundling, the final output size has reduced to 20 KiB: [bundlejs.com](https://bundlejs.com/?q=jsx-ast-utils&config=%7B%22alias%22%3A%7B%22array-buffer-byte-length%22%3A%22%40nolyfill%2Farray-buffer-byte-length%22%2C%22array-includes%22%3A%22%40nolyfill%2Farray-includes%22%2C%22array.from%22%3A%22%40nolyfill%2Farray.from%22%2C%22array.prototype.at%22%3A%22%40nolyfill%2Farray.prototype.at%22%2C%22array.prototype.find%22%3A%22%40nolyfill%2Farray.prototype.find%22%2C%22array.prototype.findlast%22%3A%22%40nolyfill%2Farray.prototype.findlast%22%2C%22array.prototype.findlastindex%22%3A%22%40nolyfill%2Farray.prototype.findlastindex%22%2C%22array.prototype.flat%22%3A%22%40nolyfill%2Farray.prototype.flat%22%2C%22array.prototype.flatmap%22%3A%22%40nolyfill%2Farray.prototype.flatmap%22%2C%22array.prototype.reduce%22%3A%22%40nolyfill%2Farray.prototype.reduce%22%2C%22array.prototype.tosorted%22%3A%22%40nolyfill%2Farray.prototype.tosorted%22%2C%22arraybuffer.prototype.slice%22%3A%22%40nolyfill%2Farraybuffer.prototype.slice%22%2C%22asynciterator.prototype%22%3A%22%40nolyfill%2Fasynciterator.prototype%22%2C%22available-typed-arrays%22%3A%22%40nolyfill%2Favailable-typed-arrays%22%2C%22deep-equal%22%3A%22%40nolyfill%2Fdeep-equal%22%2C%22define-properties%22%3A%22%40nolyfill%2Fdefine-properties%22%2C%22es-aggregate-error%22%3A%22%40nolyfill%2Fes-aggregate-error%22%2C%22es-iterator-helpers%22%3A%22%40nolyfill%2Fes-iterator-helpers%22%2C%22es-set-tostringtag%22%3A%22%40nolyfill%2Fes-set-tostringtag%22%2C%22function-bind%22%3A%22%40nolyfill%2Ffunction-bind%22%2C%22function.prototype.name%22%3A%22%40nolyfill%2Ffunction.prototype.name%22%2C%22get-symbol-description%22%3A%22%40nolyfill%2Fget-symbol-description%22%2C%22globalthis%22%3A%22%40nolyfill%2Fglobalthis%22%2C%22gopd%22%3A%22%40nolyfill%2Fgopd%22%2C%22harmony-reflect%22%3A%22%40nolyfill%2Fharmony-reflect%22%2C%22has%22%3A%22%40nolyfill%2Fhas%22%2C%22has-property-descriptors%22%3A%22%40nolyfill%2Fhas-property-descriptors%22%2C%22has-proto%22%3A%22%40nolyfill%2Fhas-proto%22%2C%22has-symbols%22%3A%22%40nolyfill%2Fhas-symbols%22%2C%22has-tostringtag%22%3A%22%40nolyfill%2Fhas-tostringtag%22%2C%22internal-slot%22%3A%22%40nolyfill%2Finternal-slot%22%2C%22is-arguments%22%3A%22%40nolyfill%2Fis-arguments%22%2C%22is-array-buffer%22%3A%22%40nolyfill%2Fis-array-buffer%22%2C%22is-date-object%22%3A%22%40nolyfill%2Fis-date-object%22%2C%22is-generator-function%22%3A%22%40nolyfill%2Fis-generator-function%22%2C%22is-regex%22%3A%22%40nolyfill%2Fis-regex%22%2C%22is-shared-array-buffer%22%3A%22%40nolyfill%2Fis-shared-array-buffer%22%2C%22is-string%22%3A%22%40nolyfill%2Fis-string%22%2C%22is-symbol%22%3A%22%40nolyfill%2Fis-symbol%22%2C%22is-weakref%22%3A%22%40nolyfill%2Fis-weakref%22%2C%22iterator.prototype%22%3A%22%40nolyfill%2Fiterator.prototype%22%2C%22object-is%22%3A%22%40nolyfill%2Fobject-is%22%2C%22object-keys%22%3A%22%40nolyfill%2Fobject-keys%22%2C%22object.assign%22%3A%22%40nolyfill%2Fobject.assign%22%2C%22object.entries%22%3A%22%40nolyfill%2Fobject.entries%22%2C%22object.fromentries%22%3A%22%40nolyfill%2Fobject.fromentries%22%2C%22object.getownpropertydescriptors%22%3A%22%40nolyfill%2Fobject.getownpropertydescriptors%22%2C%22object.groupby%22%3A%22%40nolyfill%2Fobject.groupby%22%2C%22object.hasown%22%3A%22%40nolyfill%2Fobject.hasown%22%2C%22object.values%22%3A%22%40nolyfill%2Fobject.values%22%2C%22promise.allsettled%22%3A%22%40nolyfill%2Fpromise.allsettled%22%2C%22promise.any%22%3A%22%40nolyfill%2Fpromise.any%22%2C%22reflect.getprototypeof%22%3A%22%40nolyfill%2Freflect.getprototypeof%22%2C%22reflect.ownkeys%22%3A%22%40nolyfill%2Freflect.ownkeys%22%2C%22regexp.prototype.flags%22%3A%22%40nolyfill%2Fregexp.prototype.flags%22%2C%22safe-array-concat%22%3A%22%40nolyfill%2Fsafe-array-concat%22%2C%22safe-regex-test%22%3A%22%40nolyfill%2Fsafe-regex-test%22%2C%22side-channel%22%3A%22%40nolyfill%2Fside-channel%22%2C%22string.prototype.at%22%3A%22%40nolyfill%2Fstring.prototype.at%22%2C%22string.prototype.matchall%22%3A%22%40nolyfill%2Fstring.prototype.matchall%22%2C%22string.prototype.padend%22%3A%22%40nolyfill%2Fstring.prototype.padend%22%2C%22string.prototype.padstart%22%3A%22%40nolyfill%2Fstring.prototype.padstart%22%2C%22string.prototype.replaceall%22%3A%22%40nolyfill%2Fstring.prototype.replaceall%22%2C%22string.prototype.trim%22%3A%22%40nolyfill%2Fstring.prototype.trim%22%2C%22string.prototype.trimend%22%3A%22%40nolyfill%2Fstring.prototype.trimend%22%2C%22string.prototype.trimleft%22%3A%22%40nolyfill%2Fstring.prototype.trimleft%22%2C%22string.prototype.trimright%22%3A%22%40nolyfill%2Fstring.prototype.trimright%22%2C%22string.prototype.trimstart%22%3A%22%40nolyfill%2Fstring.prototype.trimstart%22%2C%22typed-array-buffer%22%3A%22%40nolyfill%2Ftyped-array-buffer%22%2C%22typed-array-byte-length%22%3A%22%40nolyfill%2Ftyped-array-byte-length%22%2C%22typed-array-byte-offset%22%3A%22%40nolyfill%2Ftyped-array-byte-offset%22%2C%22typed-array-length%22%3A%22%40nolyfill%2Ftyped-array-length%22%2C%22unbox-primitive%22%3A%22%40nolyfill%2Funbox-primitive%22%2C%22which-boxed-primitive%22%3A%22%40nolyfill%2Fwhich-boxed-primitive%22%2C%22which-typed-array%22%3A%22%40nolyfill%2Fwhich-typed-array%22%7D%7D)

silverwind:
I'm fine with either bundling or inlining, the gains are massive in both cases. Inlining is easier and as I see it, the upsteam [rarely changes](https://github.com/jsx-eslint/jsx-ast-utils/commits/main), so it'll be low-maintenance.

antfu:
> CJS is tree-shakeable

To some extent, yes. But I think it's objective that ESM performs much better static analysis and tree-shaking. In this case, `jsx-ast-utils/elementType` is just not tree-shake very well.

> Bundle and nolyfill

I have considered that option. nolyfill is a good way to do so, but I think the main thing is that those packages are not used at all for this project.

> Sync manually with upstream

I don't think it going to be a lot of work, considering the scopes (propName and elementType) is pretty well-define and code is short. In the long term I am happy to work with @ljharb to extract as a shared and tree-shakeable utils lib if this is really a concern.

antfu:
I have merged this for now to reduce the install size for the current users. I am open to discussions for a long-term solution.

Issue #51, opened 2023-11-07
Refactor to ESM


antfu:
We could now shart refactoring js and jsx rules to ESM, for better consistency and long-term maintainability. The bundling and testing tools should already support CJS/ESM interchangeability, so we should be able to refactor them progressively. After all modules are migrated to ESM, then we could start migrating to TS later.

If you want to help, leave a comment below describing what you want to work on to avoid duplicate of work. Small PRs (targeting the [`refactor/esm`](https://github.com/eslint-stylistic/eslint-stylistic/tree/refactor/esm) branch) that migrate part by part would be appreciated instead of a single giant PR. Thanks!

PR #54, opened 2023-11-07
refactor(eslint-plugin-js): ESM migration `array-bracket-newline`

kevintyj:
Related to: #51 

## What changed?
Migrates to ESM for the rule `array-bracket-newline` under eslint-plugin-js

### DIFF
--- a/packages/eslint-plugin-js/rules/array-bracket-newline/array-bracket-newline.js
+++ b/packages/eslint-plugin-js/rules/array-bracket-newline/array-bracket-newline.js
@@ -102,8 +102,10 @@ export default {
      */
     function normalizeOptions(options) {
       const value = normalizeOptionValue(options)
+
       return { ArrayExpression: value, ArrayPattern: value }
     }
+
     /**
      * Reports that there shouldn't be a linebreak after the first token
      * @param {ASTNode} node The node to report in the event of an error.
@@ -117,8 +119,10 @@ export default {
         messageId: 'unexpectedOpeningLinebreak',
         fix(fixer) {
           const nextToken = sourceCode.getTokenAfter(token, { includeComments: true })
+
           if (astUtils.isCommentToken(nextToken))
             return null
+
           return fixer.removeRange([token.range[1], nextToken.range[0]])
         },
       })
@@ -137,8 +141,10 @@ export default {
         messageId: 'unexpectedClosingLinebreak',
         fix(fixer) {
           const previousToken = sourceCode.getTokenBefore(token, { includeComments: true })
+
           if (astUtils.isCommentToken(previousToken))
             return null
+
           return fixer.removeRange([previousToken.range[1], token.range[0]])
         },
       })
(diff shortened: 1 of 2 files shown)


kevintyj:
Was not sure if this was the type of refactor @antfu was looking for, will wait for PR review for additional changes!

packages/eslint-plugin-js/rules/array-bracket-newline/array-bracket-newline.js
```diff
@@ -3,256 +3,238 @@
...
 
 // ------------------------------------------------------------------------------
 // Rule Definition
 // ------------------------------------------------------------------------------
 
 /** @type {import('eslint').Rule.RuleModule} */
-module.exports = {
-  meta: {
-    type: 'layout',
+export const meta = {
```
ota-meshi, line 13: I don't think it's necessary to export `meta` and `create`.
I think it is enough to just export the rules to `default`. `ts` rules only have `export default`. What do you think?
https://github.com/eslint-stylistic/eslint-stylistic/blob/0a77e190c8f74145118c3ad7125637de05c63adb/packages/eslint-plugin-ts/rules/block-spacing/block-spacing.ts#L16

kevintyj: I could also just locally scope `meta` and `create` since it's being exported anyway as `default` on line 237. And if you check the tester file, it anyways consumes the module as `rule` on line 11, so I just kept the behavior from CJS over to ESM. 

Or do you mean to take the route the the TS implementation took with utils for creating rules and also the meta information as well like such: 
```
/* Check RuleCreator.d.ts from @typescript-eslint/utils */
export default createRule({
  // New information explicitly defined using the new method
  name: 'array-bracket-newline',
  meta: {
    // Same meta infomration from previous JS file
    type: 'layout',
    docs: {
      description: 'Enforce linebreaks after opening and before closing array brackets',
      recommended: false,
      url: 'https://eslint.style/rules/js/array-bracket-newline',
    },
    fixable: 'whitespace',
    // Note baseRule received from getESLintCoreRule in ts utils
    hasSuggestions: baseRule.meta.hasSuggestions,
    schema: baseRule.meta.schema,
    messages: baseRule.meta.messages,
  },

  create() {
    /* Implementation */
    return {
      ArrayPattern: check,
      ArrayExpression: check,
    }
  },
})
```

kevintyj: Only downside of doing such approach is then people would not be able to selectively import the `meta` information like `import { meta } from './array-bracket-newline'` if they are looking to extend certain parts of the code in the future.

ota-meshi: I think you can reduce the difference by `export default { meta: {...}, create () {...}}`.
Now that indentation has changed significantly, I think reviewers need to spend a lot of time.
If you want to do another refactoring beyond making it ESM, it may be better to make it a separate PR.

kevintyj: Latest commit reflect the points above and maintained as little changed code as possible for easier review :)

PR #59, opened 2023-11-10
refactor(eslint-plugin-jsx): Update jsx rules

kevintyj:
Related to: #51 #58 

## What changed?

Migrates to ESM for rules related to `eslint-plugin-jsx`

Utils are left untouched


kevintyj:
Since modules run in strict mode was wondering if `use-strict` is still required on top of each file?
Also removed `// Requirements` comment from test files as per d5159d547cb8bd37d564bce857e689b6ee2052f7 in #58

Issue #52, opened 2023-11-07
[bug] indent + brace-style 'allman': TypeScript enum opening brace is indented


PindaPixel:
When using:
```json
"rules": {
    "@stylistic/brace-style": ["warn", "allman"],
    "@stylistic/indent": ["warn", indent_size],
}
```

I expect the following to be correct:
```ts
export enum MyEnum
{
    First,
    Second,
}
```

Instead, the following occurs:

```ts
export enum MyEnum 
{
// ^ Expected indentation of 4 spaces but found 0.
    First,
    Second,
}
```

Eugeno:
Looks like @stylistic/indent for TS is broken, try to use @stylistic/js/indent
https://github.com/eslint-stylistic/eslint-stylistic/issues/46

PindaPixel:
> Looks like @stylistic/indent for TS is broken, try to use @stylistic/js/indent #46

Thanks, that works better, but not perfect. It now doesn't have any opinion on the indentation of the brace. Good enough for now, though.

PindaPixel:
https://github.com/eslint-stylistic/eslint-stylistic/pull/261

Issue #46, opened 2023-11-06
@stylistic/indent doesn't work correctly


Eugeno:
Hi, tried migrating to @eslint-stylistic, replaced just one rule, and got over a hundred errors. For example, in places like this:
![image](https://github.com/eslint-stylistic/eslint-stylistic/assets/23382920/223d8701-b951-4553-a4ee-612050d3dafe)
![image](https://github.com/eslint-stylistic/eslint-stylistic/assets/23382920/1e5b673b-fb0a-4798-91df-54fefa3120aa)

```json
"@stylistic/eslint-plugin": "1.0.0",
"eslint": "8.53.0",
```
```json
"plugins": ["@stylistic"],
"rules": {
  "@stylistic/indent": ["error", 2],
}
```

ota-meshi:
I think the cause is that you replaced it with an indent rule that TypeScript understands. If you want to use the same indent rule as ESLint, you should use the `@stylistic/js/indent` rule.

https://eslint.style/guide/migration#manual-migrate

Eugeno:
@ota-meshi thank you, this really helped! I've opened an issue because what initially confused me was the desired output shown in the first example:
```
constructor(
firstParam: string,
  secondParama: string,
)
```
As far as I checked, it's originally an issue with the `typescript-eslint` package.

ota-meshi:
As far as I know, I think typescript-eslint's indent rules stopped being maintained a long time ago.
https://github.com/typescript-eslint/typescript-eslint/issues/1824
If you want to improve the indent rule, I think they need maintenance by the community.

Eugeno:
It seems like in the @stylistic/indent config (https://github.com/eslint-stylistic/eslint-stylistic/blob/main/packages/eslint-plugin/src/index.ts), the rule for ts is overriding the corresponding rule for js. Perhaps we can simply remove it, considering that the rule for js is working perfectly fine? :) Alternatively, we could explicitly specify in that file 'indent: js.rules.indent' until the ts rule is fixed.

ota-meshi:
Is it because the indent rule don't work well with ts that you reopened this issue?
If so, I think it's better to open a new issue for each individual problem so that it's easier for someone to work on it. What do you think?

PR #60, opened 2023-11-10
Enforce the use of indent rule from the js rule set

Eugeno:
Considering that the `indent` rule for `ts` is broken (https://github.com/typescript-eslint/typescript-eslint/issues/1824), I suggest using the rule from `js` in the general default config until the broken rule is fixed or removed from the package.
related: https://github.com/eslint-stylistic/eslint-stylistic/issues/46 https://github.com/eslint-stylistic/eslint-stylistic/issues/52

### DIFF
--- a/packages/eslint-plugin/src/index.ts
+++ b/packages/eslint-plugin/src/index.ts
@@ -7,5 +7,6 @@ export default {
     ...js.rules,
     ...jsx.rules,
     ...ts.rules,
+    indent: js.rules.indent,
   },
 }


antfu:
I am not sure about it. I'd suggest you send PRs adding failing tests to the indent rules and we could work on a fix. Force to use the js rule would make many TS syntaxes uncovered by the rule. I would rather let it fail to surface the error (so we can fix it), rather that shallow it.

antfu:
I am closing this for now. Feel free to bring another PR with failing tests so we could see how we can fix it. Thanks!

PR #63, opened 2023-11-11
feat: rename `func-call-spacing` to `function-call-spacing`, provide back alias

### LINKED ISSUE #39: Rename func-call-spacing to function-call-spacing

Namchee:
## Overview

Closes #39 

This pull request adds a new alias of `func-call-spacing` named `function-call-spacing` which doesn't change the rule behavior at all.

This pull request takes a different approach from #40. Instead of renaming the rule that potentially breaks existing rule setup, it adds a new alias to the same rule.


antfu:
Did you do that manually? Many files are auto-generated. It's better to maintain an alias list in `scripts/prepare.ts` and update the script to generate them automatically.

Namchee:
@antfu, regrettably yes, since I haven't read the scripts carefully. I've added aliasing logic in the preparation script.

Namchee:
Looks like the `prepare` script is changed, will update soon

antfu (approved):
Well done. Thanks for the high-quality PR 👍

PR #66, opened 2023-11-12
feat: partial migrate to ts

antfu:
If you want to help, send PRs to `refactor/ts` branch. Thanks!


antfu:
The rules are not fully migrated yet, but since this actually improved the infra quite a lot, I think it's better to have it for now and migrate rules step by step to move faster. We have tests to cover that we don't change the runtime behaviour, I think it's pretty safe to do so.

Issue #70, opened 2023-11-13
Migrate to TypeScript


antfu:
https://github.com/eslint-stylistic/eslint-stylistic/pull/66 has done the initial work and tooling setup to bundle and type checking.

We still have many rules rewritten in JavaScript. Help would be much appreciated!

If you want to help, please feel free to leave a comment below to say what rule you want to migrate, and send PRs to the main branch, **one PR for a rule**.

Migration examples:

- Rules of `eslint-plugin-js`: https://github.com/eslint-stylistic/eslint-stylistic/commit/a88038ef8dcd854c79697381587a8932864b645c#diff-1c7f529705777d91ed7afa4e5549f6aebc607177bf49bd9f7c19d6c614defd4d
- Rules of `eslint-plugin-jsx`: https://github.com/eslint-stylistic/eslint-stylistic/commit/a88038ef8dcd854c79697381587a8932864b645c#diff-f2ed5fbe59b664a37dd073cd140a7dd6812bc40757dd51f84a6fc3407becd0a7

Namchee:
I think we need a rule tracker for this. Below are the list of rules separated by its packages. Checked rules indicates the the rule has been successfully migrated to TypeScript.

## `@stylistic/js`

- [x] `array-bracket-newline`
- [ ] `array-bracket-spacing`
- [ ] `array-element-newline`
- [ ] `arrow-parens`
- [ ] `arrow-spacing`
- [ ] `block-spacing`
- [ ] `brace-style`
- [ ] `comma-dangle`
- [ ] `comma-spacing`
- [ ] `comma-style`
- [ ] `computed-property-spacing`
- [x] `dot-location`
- [ ] `eol-last`
- [ ] `func-call-spacing`
- [ ] `function-call-argument-newline`
- [ ] `function-call-spacing`
- [ ] `function-paren-newline`
- [ ] `generator-star-spacing`
- [ ] `implicit-arrow-linebreak`
- [ ] `indent`
- [ ] `jsx-quotes`
- [ ] `key-spacing`
- [ ] `keyword-spacing`
- [ ] `linebreak-style`
- [ ] `lines-around-comment`
- [ ] `lines-between-class-members`
- [ ] `max-len`
- [ ] `max-statements-per-line`
- [ ] `multiline-ternary`
- [ ] `new-parens`
- [ ] `newline-per-chained-call`
- [ ] `no-confusing-arrow`
- [ ] `no-extra-parens`
[...]

PR #114, opened 2023-11-15
refactor(js): migrate lines-around-comment to ts

albizures:
Part of #70


packages/eslint-plugin-js/rules/lines-around-comment/lines-around-comment.ts
```diff
@@ -3,7 +3,11 @@
  * @author Jamund Ferguson
  */
 
+import { AST_NODE_TYPES } from '@typescript-eslint/utils'
```
antfu, line 6: We need to avoid importing it **at runtime**, as we only deps `@typescript-eslint/utils` on developement.

packages/eslint-plugin-js/rules/lines-around-comment/lines-around-comment.ts
```diff
@@ -261,13 +265,13 @@ export default {
...
      */
-    function isCommentAtBlockStart(token) {
+    function isCommentAtBlockStart(token: Token) {
       return (
-        isCommentAtParentStart(token, 'ClassBody')
-                || isCommentAtParentStart(token, 'BlockStatement')
-                || isCommentAtParentStart(token, 'StaticBlock')
-                || isCommentAtParentStart(token, 'SwitchCase')
-                || isCommentAtParentStart(token, 'SwitchStatement')
+        isCommentAtParentStart(token, AST_NODE_TYPES.ClassBody)
```
antfu, line 270: It's OK to keep string literals here as long as the arg is typed

PR #127, opened 2023-11-15
refactor(js): migrate indent to ts

antfu:


packages/eslint-plugin-js/rules/indent/indent.test.ts
```diff
@@ -17,16 +17,25 @@ import rule from './indent'
...
+  type: string
+  line: number
+}
+
 /**
  * Create error message object for failure cases with a single 'found' indentation type
- * @param {string} providedIndentType indent type of string or tab
- * @param {Array} providedErrors error info
- * @returns {object} returns the error messages collection
- * @private
```
kecrily, line 25: Removing the types from the JSDoc is enough. Keeping comments should be better

antfu: I don't see them as very helpful, as the variable names are already self-explaining. I don't usually remove them but since this functions has a bit weird overloads, I feel it's better to remove them instead of duplicate those messages.

PR #141, opened 2023-11-16
refactor(jsx): migrate jsx-closing-bracket-location to ts

alexzhang1030:
Part of #70


packages/eslint-plugin-jsx/rules/jsx-closing-bracket-location/jsx-closing-bracket-location.ts
```diff
@@ -3,10 +3,14 @@
...
+import { createRule } from '../../utils/createRule'
 import { docsUrl } from '../../utils/docsUrl'
-import report from '../../utils/report'
+import type { Tree } from '../../utils/types'
+import type { MessageIds, RuleOptions } from './types'
 
-const has = (obj, prop) => Object.prototype.hasOwnProperty.call(obj, prop)
+type InferSchemaByKey<K extends string> = Extract<RuleOptions[0], Record<K, any> | Partial<Record<K, any>>>
+
+const has = <const K extends string>(obj: object, prop: K): obj is InferSchemaByKey<K> => Object.prototype.hasOwnProperty.call(obj, prop)
```
antfu, line 13: It would be better if we could remove this inline util

packages/eslint-plugin-jsx/rules/jsx-closing-bracket-location/jsx-closing-bracket-location.ts
```diff
@@ -185,25 +212,35 @@ export default {
...
 
+    interface LastPropLocation {
+      column: number
+      firstLine: number
+      lastLine: number
+    }
+
+    type LastProp = LastPropLocation | undefined | Tree.JSXOpeningElement['attributes'][number]
+
+    type Tokens = ReturnType<typeof getTokensLocations>
```
antfu, line 223: Can we have those types at the top level? Thanks

alexzhang1030: That might need to extract `getTokensLocation` to the top level.

The type of `Tokens` is too complex to handwrite.

packages/eslint-plugin-jsx/rules/jsx-closing-bracket-location/jsx-closing-bracket-location.ts
alexzhang1030: We need to use this function to narrow the type, otherwise need to assert the type many times.

alexzhang1030:
And sadly about the `test coverage`, If return `true` will not pass the type check, if return `null` will not pass the `test coverage` check.

antfu (approved):

Issue #136, opened 2023-11-16
Additional package for new rules


antfu:
To make my personal day-to-day ESLint experience better, I created my personal plugin https://github.com/antfu/eslint-plugin-antfu to handle some edge cases.

Some formatting rules are quite straightforward but never presented in the ESLint/typescript-eslint rules. For example:

- [`generic-spacing`](https://github.com/antfu/eslint-plugin-antfu/blob/77b898f7a2d448e803d9eac4278bb4edeabf8364/src/rules/generic-spacing.test.ts)
- [`named-tuple-spacing`](https://github.com/antfu/eslint-plugin-antfu/blob/ce48a09f9495e9ef14d4ddc7c89ceaea239233d7/src/rules/named-tuple-spacing.test.ts)
- [`indent-binary-ops`](https://github.com/antfu/eslint-plugin-antfu/blob/b7399c27b277e91ae69e3654906ead405ff31294/src/rules/indent-binary-ops.md) - (It's quite useful!)

I assume it might be because both projects have [frozen the stylistic rules since 2020](https://eslint.org/blog/2020/05/changes-to-rules-policies/#what-are-the-changes). Now with ESLint Stylistic, the door finally open up a bit.

I want to keep the current `js` `ts` `jsx` packages as-is to keep the "1:1" migration mapping in v1.0. So I imagine we could have a separate package like `@stylistic/eslint-plugin-extra` to host new rules. Later we plan to merge all those 4 packages into a single one, as currently exposed at `@stylistic/eslint-plugin`.

While I do NOT want to allow any rules to come in and eventually burn out by maintenance burden - so even we are opening the door for new rules, we will be very strict about them. In principle, we'd suggest releasing the proposed rules as separate eslint-plugin to verify them in various codebases first. And later we could start discussing about moving them into this project.

Namchee:
Will this be included as pre-generated configs? I'm thinking of something like `plus-flat` and `plus-legacy`.

antfu:
Yes, when a new rule is added, they can add the config to the preset. I don't think we would need new variations of `plus-*`, but could directly go `recommended-*`

PR #144, opened 2023-11-16
feat: introduce `@stylistic/eslint-plugin-plus`

antfu:
#136


ljharb:
Why "plus"?

antfu:
Any name suggestions?

antfu:
A few in my mind:

- `extra` - I guess it's ok, but it feels a bit "excessive" to me.
- `additional` - I personally feel it's a bit too long.
- `more` - "JavaScript Rules, TypeScript Rules, More Rules" feels a bit weird
- `plus` - Short, with the meaning of `additional` but not as "excessive" as `extra`


In that regard, I am not a native speaker so my feeling could be inaccurate. Open for suggestions

antfu:
I am merging this to `dev` branch for now so we can start work on rules. Still open for better naming until it get published in the next minor

ljharb:
I’m a bit confused why it needs to be a separate plugin tbh. If it doesn’t belong in the main plugin, why is it in stylistic at all?

antfu:
I explained the reason in https://github.com/eslint-stylistic/eslint-stylistic/issues/136

For users who directly use `@stylistic/eslint-plugin`, it's not a concern. And eventually, we will merge all rules into a single plugin, but that would be a breaking change in the future.

ljharb:
Adding rules is semver-minor; it's only adding enabled things to a config that'd be a breaking change.

I don't find that reason compelling tbh; it just kind of seems like you're trying to move your personal rules under this quasi-official branding :-/

antfu:
> you're trying to move your personal rules under this quasi-official branding :-/

I found that offensive if you say so. I spend a lot of effort on migrating and maintaining this because I care. I tried hard to keep things natural, and only proposed to add rules that improve whitespace and solves long-standing issue that no one was going to fix. I don't know how I can help communicate this better. You ignore the technical reasoning and accuse me like that out of nowhere.

ljharb:
I don't mean to make an accusation, but you're making a ton of unilateral changes without many others weighing in.

PR #145, opened 2023-11-16
refactor(jsx): migrate jsx-max-props-per-line to ts

alexzhang1030:
Part of #70


antfu:
It seems you forget to remove the js file?

antfu (approved):

PR #150, opened 2023-11-16
refactor(jsx): migrate jsx-curly-brace-presence to ts

alexzhang1030:
Part of #70

PR #151, opened 2023-11-16
refactor(jsx): migrate jsx-props-no-multi-spaces to ts

alexzhang1030:
Part of #70,

Some code has many type assertions, that's because the type narrowing does not work as expected.


alexzhang1030:
same with #150, test cases are not typed.

Issue #45, opened 2023-11-05
sourceCode undefined in every rule


Silon:
I created package with own rules where I use stylistic with next/recommended

```js
// eslin-plugin-test/index.js

module.exports = {
  configs: {
    nextjs: {
      plugins: ["@stylistic"],
      extends: ["plugin:@next/next/recommended"],
      rules: {
        "@stylistic/max-len": ["error", 120],
      },
    },
  },
};
```

Then I use it in my NextJS project

```js
//  nextproject/.eslintrc.js

/** @type {import('eslint').Linter.Config} */
module.exports = {
  root: true,
  extends: ['plugin:@test/test/nextjs'],
  parserOptions: {
    project: './tsconfig.json',
    ecmaVersion: 'latest',
  },
  env: {
    es6: true,
    node: true,
  },
};
```

The problem is that every rule added into rules in index.js file causes an error:

```
[Error - 18:41:35] TypeError: Cannot read properties of undefined (reading 'lines')
Occurred while linting /Users/silon/Developer/nextjs-template/.eslintrc.js:2
Rule: "@stylistic/max-len"
    at checkProgramForMaxLength
```

The problem in this case is in `max-len.js` file in this line:

```js
function checkProgramForMaxLength(node) {
  // split (honors line-ending)
  const lines = sourceCode.lines;
```

When I remove max-len rule from index.js eslint works good

ota-meshi:
Hi @Silon.
I think you need to use ESLint v8.40 or newer.
New those properties were added in v8.40.

https://eslint.org/blog/2023/05/eslint-v8.40.0-released

Silon:
@ota-meshi you're right! In my library I got newest but project had old version. Thank you. Issue might be closed

PR #172, opened 2023-11-18
docs: specify the Node engine and ESLint version requirement

antfu:
We inherited those requirements from the source projects. This PR makes it clear to users.

### DIFF
--- a/docs/guide/faq.md
+++ b/docs/guide/faq.md
@@ -18,6 +18,13 @@ According to [`typescript-eslint`'s definition](https://typescript-eslint.io/lin
 
 For ESLint Stylistic, our primary focus is on the **formatting** and **stylistic** rules inherited from `eslint` and `typescript-eslint`. We will maintain some stylistic rules; however, not all will be included. Their inclusion depends on whether the upstream projects choose to retain them. We welcome new rules proposed by the community for the future when we move into the maintenance phase and develop the infrastructure to introduce experimental rules. For more details, track the [Project Progress](/contribute/project-progress).
 
+## What's the requirements for ESLint Stylistic?
+
+Since majority of the rules are migreated from ESLint v8 and `typescript-eslint` v6, we inherit the same requirements:
+
+- Node.js >=v16.0.0
+- ESLint >=v8.0.0
+
 ## How to auto-format on save?
 
 ##### VS Code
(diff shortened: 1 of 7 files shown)


docs/guide/faq.md
```diff
@@ -18,6 +18,13 @@ According to [`typescript-eslint`'s definition](https://typescript-eslint.io/lin
 
 For ESLint Stylistic, our primary focus is on the **formatting** and **stylistic** rules inherited from `eslint` and `typescript-eslint`. We will maintain some stylistic rules; however, not all will be included. Their inclusion depends on whether the upstream projects choose to retain them. We welcome new rules proposed by the community for the future when we move into the maintenance phase and develop the infrastructure to introduce experimental rules. For more details, track the [Project Progress](/contribute/project-progress).
 
+## What's the requirements for ESLint Stylistic?
+
+Since majority of the rules are migreated from ESLint v8 and `typescript-eslint` v6, we inherit the same requirements:
```
Namchee, line 23: Typo at `migreated` here

```suggestion
Since majority of the rules are migrated from ESLint v8...
```

antfu: Why remove the last part? ESLint v8 support down to Node 14 but ts-eslint ask for 16, so we take 16

Namchee: I unknowingly omitted it, my bad. It should only replace `migreated` with `migrated`

packages/eslint-plugin-js/package.json
```diff
@@ -100,11 +100,17 @@
...
+  "engines": {
+    "node": "^16.0.0 || >=18.0.0"
+  },
   "scripts": {
     "build": "rimraf dist && rollup --config=rollup.config.mts --configPlugin=rollup-plugin-esbuild",
     "dev": "rollup --config=rollup.config.mts --configPlugin=rollup-plugin-esbuild --watch",
     "prepublishOnly": "pnpm build"
   },
+  "peerDependencies": {
+    "eslint": ">=8.0.0"
```
ota-meshi, line 112: I think we probably need the version at the time we forked it. I don't think it will work unless we use at least a newer version than v8.40.

https://github.com/eslint-stylistic/eslint-stylistic/issues/45#issuecomment-1793911075

PR #174, opened 2023-11-19
docs: unify `incorrect/correct` label for typescript documentation

ice-blaze:
Typescript documentation used emojis to display incorrect/correct code examples. Now it use the same `::: correct` tag to display the labels in the documentation.

### Description

This PR unifies the label `correct` incorrect` that was displayed on the typescript documentation. Now it uses the same label than the Javascript documentation.

### Additional context

Didn't created an issue for it since it was such a small change but if it's requested. I'll do that.


kecrily (approved):
Good catch

Issue #188, opened 2023-11-27
Bug: no-multi-spaces doesn't count tabs as space separators


otaxhu:
### Describe the bug

When you have something like this:
```js
let foo = ( 2 + 2 ); // some statement or expression that has parenthesis,
                     // I only tested out this problem with this
```
And you have the `"@stylistic/js/space-in-parens": "error"`  rule, the linter will complain about the extra spaces that has the parenthesis, and will auto-fix, everything is fine, even with tabs and spaces mixed, for example:
```js
let foo = (\t 2 + 2     \t);
```
this two examples will auto-fix to this:
```js
let foo = (2 + 2);
```
The problem comes when you also don't want to have multiple spaces between unary operators, identifiers, keywords, etc. with the `"@stylistic/js/no-multi-spaces": "error"` with this example the linter will complain about and will throw an error (so nothing wrong with this example):
```js
let foo = (2 +    2); // only spaces, no tabs
// will auto fix to: let foo = (2 + 2);
```
but with the next two example the linter will say nothing, even when it's breaking the rule for `no-multi-spaces`:
```js
// 1st example
let foo = (2\t\t\t+\t\t\t\t\t2); // with tabs

// 2nd example
let foo = (2 \t\t+\t\t 2); // with tabs and with only one space in both sides of the plus sign (+)
```
#### BUT...
If you have two spaces one just after another without any tabs in the middle it will complain and succesfully throw an error
```js
let foo = (2\t\t  +\t\t    2); // two spaces one after another in the left side of + and four spaces on the other side
// will auto-fix to: let foo = (2 + 2);
```

I'm bringing out the `"space-in-parens"` example because it succesfully treats the tabs as a separator and throws the error, but this `no-multi-spaces` doesn't.

Another example with more than one space that I found later when I was writing this issue that also don't throw an error:
```js
let foo = (2 \t \t + \t \t \t 2); // I think I found a pattern here, only throws an error if founded a space and if the next token is a space then throw, else don't
// I'm not an expert with the tokenizing process of parsers of javascript so I'm not of much help in this area.
```

### Reproduction

https://stackblitz.com/edit/eslint-stylistic-eslint-stylistic-skt16h?file=playground%2Findex.ts&view=editor

### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
[...]

PR #195, opened 2023-11-30
feat(js/no-multi-spaces): new `includeTabs` option

antfu:
fix https://github.com/eslint-stylistic/eslint-stylistic/issues/188


ota-meshi (approved):
LGTM!

packages/eslint-plugin-js/rules/no-multi-spaces/no-multi-spaces.ts
```diff
@@ -36,6 +36,10 @@ export default createRule<MessageIds, RuleOptions>({
             type: 'boolean',
             default: false,
           },
+          includeTabs: {
+            type: 'boolean',
+            default: false,
```
kecrily, line 41: I think we can turn it on by default.

PR #207, opened 2023-12-03
fix(js/function-paren-newline): prevent newline error in type parameter

### LINKED ISSUE #206: False positive in rule: function-paren-newline when using type parameter
### Validations

### Describe the bug

In several scenarios first opening paren is found in type parameters, which leads to falsy error and incorrect fix

Examples:

- function declaration

```ts
function a<A extends Array<any>, T extends (...args: any[]) => any> (
  b: T,
  c: any,
): any {}
```

- function expression

```ts
const a = function <A extends Array<any>, T extends (...args: any[]) => any> (
  b: T,
  c: any,
): any {}
```

- call expression

```ts
a<Array<any>, (...args: any[]) => any>(
  b,
  c,
)
```

### Reproduction

in description

### Contributes

- [X] I am willing to submit a PR to fix this issue
- [X] I am willing to submit a PR with failing tests

levchak0910:
First commit with failing tests, second with actual fix

fixes #206


packages/eslint-plugin-js/test-utils/parsers/function-paren-newline/call-expression-type-parameter.js
ota-meshi, line 1: I don't think we should implement parsers ourselves.  I think we should use the parsers that exist to check compatibility with the ecosystem.

levchak0910: I also had a bad feeling about it, but just followed the way how other were writing previously

Should we use the latest version of typescript-eslint/parser?

antfu:
Since this is actually a TS syntax issue, I wonder the reason you choose to fix it inside the js rule instead of ts? I am ok to have them in js rule if it would be too hard to workaround in the ts rule, as eventually we are going to merge them into one. But if possible, I think we favor having them in ts rule in order to have a clear scope.

levchak0910:
Yes, it is TS syntax, therefore I wrote the condition to ignore TS related tokens. Current implementation found first opening paren in type parameter which leads to false positive error and wrong fix. 

I mean, JS rule reports an error in TS code and fixes TS code, which is not correct - therefore the fix is here. Also, it doesn't introduce any improvements for TS code

PR #213, opened 2023-12-08
fix(ts/padding-line-between-statements): add new `function-overload` statement type

### LINKED ISSUE #190: `padding-line-between-statements` function overload support

I would like to avoid adding an empty line before function overloads and the function itself. Currently, there's no way to specify overloads as a statement in the rules options

Actual
```ts
function foo(): void;
function foo(param0?: boolean): boolean;

function foo(param0?: boolean) {
  //
}
```

### Suggested solution

Since `interface` and `type` has their own statement key, having `function-overload` or `overload` `fn-overload`(👎🏻 ) as an alias would be better.

Expected:
```ts
function foo(): void;
function foo(param0?: boolean): boolean;
function foo(param0?: boolean) {
  //
}
```

### Alternative

_No response_

### Additional context

```js
{
    '@stylistic/padding-line-between-statements': [
        'error',
        {
            blankLine: 'always',
            next: ['multiline-block-like'],
            prev: ['*'],
        },
    ]
}
```

### Validations

antfu:
close #190

### DIFF
--- a/packages/eslint-plugin-ts/rules/padding-line-between-statements/README.md
+++ b/packages/eslint-plugin-ts/rules/padding-line-between-statements/README.md
@@ -3,19 +3,22 @@ description: 'Require or disallow padding lines between statements.'
 ---
 
 This rule extends the base [`padding-line-between-statements`](/rules/js/padding-line-between-statements) rule.
+
 It adds support for TypeScript constructs such as `interface` and `type`.
 
 ## Options
 
-In addition to options provided by ESLint, `interface` and `type` can be used as statement types.
+In addition to options provided by ESLint, the following options can be used as statement types:
+
+- `interface`
+- `type`
+- `function-overload`
 
 For example, to add blank lines before interfaces and type definitions:
 
 ```jsonc
 {
-  // Example - Add blank lines before interface and type definitions.
-  "padding-line-between-statements": "off",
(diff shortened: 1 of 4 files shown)


kecrily (approved):

packages/eslint-plugin-ts/rules/padding-line-between-statements/README.md
```diff
@@ -26,4 +29,17 @@ For example, to add blank lines before interfaces and type definitions:
...
+{
+  "@stylistic/padding-line-between-statements": [
+    "error",
+    {
+      "blankLine": "never",
+      "prev": "function-overload",
+      "next": "function"
+    }
+  ]
+}
```
kecrily, line 44: I think we need to add it to the default options.

kecrily: It might be more readable that way. But that's a personal opinion, so it doesn't matter.

antfu: We could add it to the default shared config, tho. But IIRC, we didn't enable those rules at all. I guess I was afraid it would be too much and hard to balance well. If we could come up with a solid config to cover majority of the cases, I am happy to have it.

Issue #218, opened 2023-12-10
Add comma dangle option to config factory


eagerestwolf:
### Clear and concise description of the problem

I want a comma dangle option for the config factory function so I can burn all the comma dangles to the ground without doing so manually 😄.

I despise comma dangles. They are ugly, non-functional, and only serve to make code flow more difficult to understand (at times). Granted, I am well aware that some developers like them because they make expanding arrays, objects, etc. easier, and that's okay. Everyone is allowed to be wrong from time to time (I jest). But for real, being able to control things like comma dangles would be a godsend. Before anyone asks, yes I use double quotes as well (I know, I'm a terrible JS developer, go back to C# blah, blah, blah).

### Suggested solution

In the customize config, the Options could be changed to:

```ts
export interface StylisticCustomizeOptions<Flat extends boolean = true> {
  /**
   * Get a flat config
   * @default true
   */
  flat?: Flat
  /**
   * The name of the registered plugin, used to prefix rule IDs
   * @default '@stylistic'
   */
  pluginName?: string
  /**
   * Indentation level
   * @default 2
   */
  indent?: number | 'tab'
  /**
   * Quote style
   * @default 'single'
   */
  quotes?: 'single' | 'double'
  /**
   * Whether to enable semicolons
   * @default false
   */
  semi?: boolean
  /**
   * Enable JSX support
   * @default true
   */
  jsx?: boolean

  /**
   * Whether to enable comma dangles
   * @default 'always-multiline'
   */
  commaDangle?: 'never' | 'always' | 'always-multiline' | 'only-multiline'
}
```

Then the offending rule could be changed to:

```ts
export function customize(options: StylisticCustomizeOptions<boolean> = {}): Linter.FlatConfig | Linter.BaseConfig {
  const {
    flat = true,
    indent = 2,
    jsx = true,
    pluginName = '@stylistic',
    quotes = 'single',
    semi = false,
    commaDangle = 'always-multiline'
  } = options

  let rules: Rules = {
    // ...other rules
    '@stylistic/comma-dangle': ['error', commaDangle],
    // ...yet more rules
  }

  // ...rest of function (I'm feeling lazy, okay?)
```

### Alternative

I mean, I could just manually override the rule, but that's kind of a pain because I need to deep merge the customize config with my rule modification, just to override 1 rule, or I have to do all of the rules manually. Neither of which are really appealing options.

### Additional context

[...]

PR #224, opened 2023-12-12
feat(configs): add new options to customize config to assist with migrating from Prettier

eagerestwolf:
### Description

This PR adds support for more Prettier options to the customize config to allow users to transition from using Prettier to using eslint-stylistic. In total, 4 new options where added to the customize config.

This PR also adds documentation for users coming from Prettier on using stylistic-eslint as a substitute for Prettier.

### Options already present

- `indent` (this covers Prettier's `tabWidth` and `useTabs`)
- `quotes`
- `semi`

### Options added

- `arrowParens`
- `blockSpacing` (roughly equivalent to Prettier's `bracketSpacing`)
- `quoteProps`
- `commaDangle` (equivalent to Prettier's `trailingCommas`)

#### Because JSX

I am not really a JSX developer (I'm more of a svelte guy), so I didn't mess with anything relating to JSX. This includes the options listed below:

- `jsxSingleQuote`
- `bracketSameLine`
- `htmlWhitespaceSensitivity`
- `singleAttributePerLine`

#### Because parsing

The following options weren't added either because they serve no purpose under ESLint; they simply defeat the purpose of linting to begin with; or ESLint already has an equivalent built-in

- `range`
- `parser`
- `requirePragma`
- `insertPragma`
- `embeddedLanguageFormatting`

#### Because covered by another rule not in config

- `endOfLine` (covered by `linebreak-style`)
- `printWidth` (covered by `max-len`)
- `proseWrap` (also covered by `max-len`)

### Linked Issues

#218 

### Additional context

I apologize if my documentation if overly verbose, I am just a fan of clear, highly explanatory documentation. Also, I did my best to match the tone of the rest of the documentation site to hopefully make it less obvious that someone else wrote it.

The big issue at the moment, as referenced during the discussion in #218, is adding 4 options too many? Also, it should be noted that there are 4 more options still not covered (as of right now). Ultimately, I leave that issue in the hands of the maintainers at this point. Also, feel free to squash and commit when the time comes, I just followed conventional commits and updated the title to reflect the conventional commit standard as well.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -20,6 +20,7 @@ const GUIDES: DefaultTheme.NavItemWithLink[] = [
   { text: 'Why', link: '/guide/why' },
   { text: 'Shared Configs', link: '/guide/config-presets' },
   { text: 'Migration', link: '/guide/migration' },
+  { text: 'Migration from Prettier', link: '/guide/migration-from-prettier' },
   { text: 'FAQ', link: '/guide/faq' },
 ]
 
--- a/docs/guide/migration-from-prettier.md
+++ b/docs/guide/migration-from-prettier.md
@@ -1,0 +1,165 @@
+# Migrating from Prettier
+
+Migrating from using Prettier for code formatting to using this project is extremely straightforward, and this page will walk you through the process of getting set up. For more information as to why you might use this package in place of Prettier, refer to [Why](/guide/why).
+
+The first step is to follow the [Getting Started](/guide/getting-started) guide. This will walk you through the process of getting setup to use this package, as well as adding it to your config.
+
+## Options
+
+### Supported Options
+
+The following Prettier options are supported:
+
+- `arrowParens`
+  - default: `false`
+  - supported values:
+    - `true`: always use parenthesis around arrow functions
+    - `false`: avoid parenthesis around arrow functions
+- `bracketSpacing` (use option `blockSpacing`)
+  - default: `true`
+  - supported values:
+    - `true`: always use spaces
+    - `false`: never use spaces
+- `endOfLine` (see [printWidth and endOfLine](#printwidth-and-endofline))
+- `printWidth` (see [printWidth and endOfLine](#printwidth-and-endofline))
+- `quoteProps`
+  - default: `'consistent-as-needed'`
+  - supported values:
+    - `'always'`: always quote object properties, whether required or not
+    - `'as-needed'`: only quote object properties that require quotes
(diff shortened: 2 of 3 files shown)


antfu:
I am good with adding those new options but I am not so sure about the guide for "Migration from Prettier". I am not even sure if it's something we should actively advocate. Since there is so much to trade off on choosing Prettier or not and we proprobably won't match the feature parity. I personally think it would be better to not have that page, but instead, have each options' JSDocs mention it like:

```ts
{
  /**
   * When to enable arrow parenthesis
   *
   * Similar to the `arrowParens` in Prettier
   * @default false
   */
  arrowParens?: boolean
}
```

To make it useful while been a bit more passive.

eagerestwolf:
Yeah, mixing Prettier and ESLint while explicitly not using Prettier on code is certainly not an intended use case of Prettier for sure. If I'm honest, the documentation actually already (in a roundabout way) does advocate for replacing Prettier with ESLint (namely, the Why page). However, I will concede that this plugin will certainly never attain feature parity with Prettier simply due to the fact that ESLint was never intended to be a formatter (hence spinning style rules off into a plugin). With that said, I think that compromising by documenting the Prettier feature equivalents in JSDoc is a good compromise. I will get on that as soon as I can.

antfu:
In the why section, we explain the reason why this project is vital to our use case. Indeed, in our opinion, we would prefer not to use Prettier due to its limitations. The reasoning is more for users to judge if their use cases fit into our model rather than trying to convince everyone to move away from Prettier (I believe, in many cases, Prettier is a very good tool for their use cases). Having a "Migrate to Prettier" might indicate that we **could replace Prettier** and might also be a bit aggressive. I'd prefer to do it a bit more passively, while provide necessary information on there it's needed (e.g. in JSDocs)

antfu:
Or maybe instead of "Migrating to Prettier" we could have a more neutral narrative like "Features Partite with Prettier" and explain the difference and limitations?

eagerestwolf:
> Or maybe instead of "Migrating to Prettier" we could have a more neutral narrative like "Features Partite with Prettier" and explain the difference and limitations?

I think this could be a better approach, that way people are aware that this plugin could serve as a replacement for Prettier (if they desire). It would also be consistent with the messaging presented on the `Why?` page. For the time being, I will roll back the documentation commit (but I'll keep the changes staged somewhere else), and update the JSDoc comments as suggested. Also, sorry for the hiatus, work was crazy leading up to the holidays, and then of course there was the holidays.

eagerestwolf:
Ok, docs change has been reverted until we determine what the appropriate response should be on that front. In the meantime, I updated the JSDoc comments to reflect the Prettier options replicated.

antfu (approved):

Issue #229, opened 2023-12-14
Use Consultation


SuperCodeCandy:
I'm not sure which template to use to create my question, so I started a new template.

I have a question. When writing the following code, I encountered an error warning under my `id` and `username`, and the editor prompted me with an `Expected indication of 4 spaces but found 2.eslint@stylistic /Js/index`.

```ts
import { Entity, Column, PrimaryGeneratedColumn } from 'typeorm';

@Entity()
export class UserEntity {
  @PrimaryGeneratedColumn()
  id: string;

  @Column()
  username: string;
}
```

I'm sorry, but I couldn't find any demonstration or explanation on how to solve this error prompt after reviewing the document. I think this indentation format is an acceptable one. Because, how should I configure it to solve this problem?

antfu:
Duplicated of https://github.com/eslint-stylistic/eslint-stylistic/issues/208

PR #231, opened 2023-12-15
fix: add handler of PropertyDefinition for indent rule

### LINKED ISSUE #208: `indent` can't work correctly after PropertyDefinition
### Validations

### Describe the bug

with the rule `'@stylistic/indent': [error, 4]`, 
`indent` can't work correctly after PropertyDefinition:

Origin Code: 
```
class Foo {

    @property
age: number = 0;

}
```

 the Expected Result like this:
```
class Foo {

    @property
    age: number = 0;

}
```

But indent can't work correctly.

### Reproduction

https://stackblitz.com/edit/eslint-stylistic-eslint-stylistic-ffq7ge?file=playground%2Findex.ts,playground%2Feslint.config.js

### Contributes

Namchee:
### Description

This pull request solves problem with `indent` rule when dealing with property decorators by adding a handler for `PropertyDefinition`.

### Linked Issues

Closes #208 

### Additional context

Two new test cases are added, from #208 and #229

### DIFF
--- a/packages/eslint-plugin-ts/rules/indent/indent.ts
+++ b/packages/eslint-plugin-ts/rules/indent/indent.ts
@@ -181,6 +181,8 @@ export default createRule<RuleOptions, MessageIds>({
           rules['*:exit'](node)
       },
 
+      'PropertyDefinition': () => {},
+
       VariableDeclaration(node: Tree.VariableDeclaration) {
         // https://github.com/typescript-eslint/typescript-eslint/issues/441
         if (node.declarations.length === 0)
--- a/packages/eslint-plugin-ts/rules/indent/indent.test.ts
+++ b/packages/eslint-plugin-ts/rules/indent/indent.test.ts
@@ -1729,6 +1729,7 @@ age: number = 0;
     @property
         id: string;
 
+
     @property
     isActive: boolean;
     @property bar: boolean;
@@ -1744,6 +1745,7 @@ class Foo {
 
     @property
     id: string;
+
 
     @property
     isActive: boolean;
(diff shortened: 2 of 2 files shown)


packages/eslint-plugin-ts/rules/indent/indent.ts
```diff
@@ -181,6 +181,8 @@ export default createRule<RuleOptions, MessageIds>({
           rules['*:exit'](node)
       },
 
+      'PropertyDefinition': () => {},
```
ota-meshi, line 184: Could you please add test cases to test if it gets indented correctly if I use some line breaks inside a class field?
I'm worried that it won't work well because it overrides and erases the `PropertyDefinition` processing of the eslint-plugin-js `indent` rule.

Namchee: I have added few cases in 6ed5b09.

ota-meshi: The tests added were not what I intended. For example, break lines within the class field as below.

```js
class Foo {
    [
        x
    ] =
        [
            42
        ]
}
```

I don't think it will get indented well with the current PR changes.

Namchee: I see what you mean. Given this input

```ts
class Foo {
    [
      x
    ] = 
      [
        42
      ]
}
```

the output from this change is 

```ts
class Foo {
    [
    x
    ] = 
    [
        42
    ]
}
```

which is incorrect as `x` is missing 4 spaces, I'll try to delve deeper regarding this issue.

Namchee: After thinking through, I think it's better if we fixed it by handling `Decorator` in the JS plugin file.

orgads:
@Namchee Thanks for this PR. Any progress with it?

PR #239, opened 2023-12-22
fix(lines-between-class-members): properly infer `exceptAfterOverload` from `enforce`

jritzi:
### Description

PR with failing test for issue #240 

The `exceptAfterOverload: true` option for the TS version of lines-between-class-members only applies if the first options is `'always'`. If using the Object version with `'enforce'`, overloads are still flagged as an error.

### Linked Issues

https://github.com/eslint-stylistic/eslint-stylistic/issues/240

### Additional context


antfu:
Could you add the test case at the end of the list with a comment linking to the issue? Thanks.

jrdioko:
@antfu sure, done.

antfu (approved):

Issue #241, opened 2023-12-22
Add `spaceMode` to `jsx-one-expression-per-line`


exoRift:
### Clear and concise description of the problem

`--fix` for `jsx-one-expression-per-line` automatically puts in `{' '}` children when I want it to use `&nbsp;`

### Suggested solution

Add a `spaceMode` option to the rule for JS-style spacing or HTML-style spacing

### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Contributes

- [X] If this feature request is accepted, I am willing to submit a PR to fix this issue

antfu:
Sounds good to me, PR welcome :)

PR #248, opened 2023-12-29
feat(jsx-one-expression-per-line): add `spaceMode` option (#241)

exoRift:
### Description
Adds the `spaceMode` option to `jsx-one-expression-per-line` which uses the HTML `&nbsp;` entity instead of JSX `{' '}`

### Linked Issues
#241

There are a few issues:
- I wasn't able to generate the `types.d.ts` file (I couldn't figure out how)
- The test is only replacing the text for the first offending source and not the second offending source. I'm not sure why
- There's an issue with how this rule parses sources. If you look at the test I added, the first literal has a space preceding the embedded div. When this rule runs, it treats the preceding space as part of the first literal (which doesn't get targeted because it's not a violation) and thus ignores it. For proper fixing, the rule would have to _replace_ that space with `&nbsp;` and then add a newline, however, there's no way for me to do that and simply adding `&nbsp;` and then a newline creates two spaces. I'm not sure how to resolve this issue
  - In fact, this is an issue with the rule itself even without these changes as the fix keeps that space and then adds a newline with `{' '}` as well. (Which, again, is two spaces)

### DIFF
--- a/packages/eslint-plugin-jsx/rules/jsx-one-expression-per-line/jsx-one-expression-per-line.test.ts
+++ b/packages/eslint-plugin-jsx/rules/jsx-one-expression-per-line/jsx-one-expression-per-line.test.ts
@@ -1558,5 +1558,31 @@ Go to page 2
       ],
       parserOptions,
     },
+    {
+      code: `
+        <div>
+          literal <div /> second literal
+        </div>
+      `,
+      options: [{ spaceMode: 'html' }],
+      output: `
+        <div>
+          literal&nbsp;
+<div />
+&nbsp;second literal
+        </div>
+      `,
+      errors: [
+        {
+          messageId: 'moveToNewLine',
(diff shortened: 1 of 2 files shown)


antfu:
To answer your questions:

1. You can run `npm run update` on the project root
2. That's how rules works, only one change is applied at one time. You could have some basic test for one occurrence (and check the error messages), and then create a new set of tests like https://github.com/eslint-stylistic/eslint-stylistic/blob/59070ae45258046e88bb4bcc36cbd74f86eec9bf/packages/eslint-plugin-plus/rules/indent-binary-ops/indent-binary-ops.test.ts#L31-L55 to test the final result.

so1ve:
@exoRift Are you on windows? I met this error: https://github.com/privatenumber/tsx/issues/345 and I had to use WSL to develop

PR #271, opened 2024-01-26
feat(jsx/pascal-case): new rule port from eslint-plugin-react

SukkaW:
### Description

The PR ports the `jsx-pascal-case` rule from `eslint-plugin-react` to `eslint-stylistic`.

React and Solid.js only enforce components' names to be capitalized. Enforcing components' names to be Pascal-case is more of a stylistic rule.

### Linked Issues

N/A

### Additional context

Rule and tests are ported from `eslint-plugin-react` directly, with only a few changes:

- Switch to eslint stylistic's built-in utility for `isDOMComponent` and `getElementType`
- Minor change to the rule option schema (`ignore` should be `string[]` instead of `[] | [string]`)
- Prefer the restructure syntax when possible (Since we are not targeting Node.js 4, right?)
- Replace `lastIndexOf() > -1` with `includes` (Since we are not targeting Node.js 4, right?)
- Replace `minimatch` with `picomatch` (They behave almost identical, but `picomatch` is way lighter)
  - After replacement, all original tests still pass
  - The PR also improves performance by hoisting the picomatch's matcher, which prevents re-creation

The changes passed lint and test cases on my machine locally.


packages/eslint-plugin-jsx/rules/jsx-pascal-case/README.md
```diff
@@ -0,0 +1,75 @@
...
+</TestComponent>
+
+<CSSTransitionGroup />
+```
+
+## Rule Options
+
+```js
+...
+"react/jsx-pascal-case": [<enabled>, { allowAllCaps: <boolean>, allowNamespace: <boolean>, allowLeadingUnderscore: <boolean>, ignore: <string[]> }]
```
kecrily, line 33:
```suggestion
"@stylistic/jsx/jsx-pascal-case": [<enabled>, { allowAllCaps: <boolean>, allowNamespace: <boolean>, allowLeadingUnderscore: <boolean>, ignore: <string[]> }]
```

kecrily (commented):
LGTM and thanks.

kecrily (approved):

antfu (approved):

PR #273, opened 2024-01-29
docs: show red underlines in examples in rules docs

ota-meshi:
### Description

This PR implements ShikijiTransformer to indicate with a red marker where an error occurs in the example rules document.

Example:
image

### Linked Issues

None.

### Additional context

I would like to hear your opinion on displaying markers in documents.

This is my first time implementing ShikijiTransformer, so I'm still not sure if this PR will work.

I have submitted a PR to ESLint that does something similar.
https://github.com/eslint/eslint/pull/18041

I would like to hear your opinion on the marker style.


antfu:
Very cool! 🔥

antfu:
That's a very interesting idea and I think it's even deserved to be a general transformer/integration similar to TwoSlash - where it could be possible to us as a part of tests as well. I'll think about the design a bit

antfu:
Made a few changes to Shiki core to make this easier, and now we have:

image

Again, thanks for the amazing idea!

Do you expect to use this in the docs of your other plugins? If so, we could make a package for it.

antfu (approved):

PR #318, opened 2024-04-11
chore: create `dependabot.yml`

MichaelDeBoey:
Just like we have in all other repositories

This makes sure that our dependencies + actions stay up to date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
+
+  - package-ecosystem: npm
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
```
antfu, line 6: Daily might be a bit too frequent? Maybe weekly?

MichaelDeBoey: The first run might be a bit overwhelming, but once versions are up-to-date, this is fine imo

It's what we do in other repos as well

PR #337, opened 2024-04-11
feat!: requires Node `^18.18.0 || ^20.9.0 || >=21.1.0`, bump deps

antfu:
This PR bumps `@typescript-eslint/*` to v7, and `eslint` to v9. Inheritly, the Node version now requires `^18.8.0 || >=20.0.0`. We consider this a breaking change and will do as a minimal v2.0 release.

Due to the changes in ESLint v9, we need to migrate the tests.


MichaelDeBoey (changes requested):
I would suggest to use minimum 18.18 instead of 18.8 as that's what `@typescript-eslint/*` needs

packages/eslint-plugin-js/package.json
```diff
@@ -101,7 +101,7 @@
     "rules/**/*.d.ts"
   ],
   "engines": {
-    "node": "^16.0.0 || >=18.0.0"
+    "node": "^18.8.0 || >=20.0.0"
```
voxpelli, line 104:
```suggestion
    "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
```

Most of the other ESLint-Community projects seems to align with ESLint 9 and use this range instead

If you want to ensure it's not too wide you can do so automatically with my: https://www.npmjs.com/package/installed-check

voxpelli:
@antfu What remains to move past draft state?

antfu:
We are waiting for some management process to get out the last path release and then we could have it this one very soon. Probably next week would do.

antfu:
I think we will need to wait `typescript-eslint` to migrate to ESLint v9 first: https://github.com/typescript-eslint/typescript-eslint/issues/8211

voxpelli:
@antfu Looks like nothing is really blocking in typescript-eslint except for the annoyance it causes internally here in that the peer dependency range is invalid

antfu:
ESLint Stylistic itself already works with ESLint v9 today.

However, in order to move forward, the tests need to be migrated, and `@typescript-eslint/rule-tester` is still on ESLint v8, which makes the test unreliable at this moment. I am not sure how much they would need to do to migrate the RuleTester for v9. But in case the potential major version release they might need to do, I would prefer to wait until they migrate over, to avoid two major versions on our side in a short amount of time.

PR #352, opened 2024-04-17
`ts/lines-between-class-members`: does not apply to interface members

ygrandgirard:
### Description

PR with failing test for issue https://github.com/eslint-community/eslint-stylistic/issues/353.
The `ts/lines-between-class-members` rule does not apply to interface members.

Given this config:
```json
{
  "rules": {
    "@stylistic/ts/lines-between-class-members": "error"
  }
}
```

This will be reported:
```ts
class A {
  a: string;
  b: string; // Expected blank line between class members.
}
```

But not this:
```ts
interface A {
  a: string;
  b: string;
}
```

### Linked Issues
https://github.com/eslint-community/eslint-stylistic/issues/353

### Additional context
It does not look particularly difficult to fix (see https://github.com/ygrandgirard/eslint-stylistic/compare/main...ygrandgirard:eslint-stylistic:bugfix/lines-between-interface-members) though I am not sure about the typing. If you are ok, I can create another PR with my fix.

### DIFF
--- a/packages/eslint-plugin-ts/rules/lines-between-class-members/lines-between-class-members.test.ts
+++ b/packages/eslint-plugin-ts/rules/lines-between-class-members/lines-between-class-members.test.ts
@@ -129,6 +129,25 @@ abstract bar(a: string, b: string): void;
       `,
       options: ['always'],
     },
+    {
+      code: `
+interface foo {
+bar(): void;
+
+baz(): void;
+};
+      `,
+      options: ['always'],
+    },
+    {
+      code: `
+interface foo {
+bar(): void;
+baz(): void;
+};
+      `,
+      options: ['never'],
+    },
   ],
   invalid: [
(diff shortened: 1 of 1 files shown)


antfu:
I don't understand - the rule name is specifically mention it's for `class-members` - I don't think lint for interface making sense.

ygrandgirard:
The way I see TypeScript interfaces is as syntactic sugar around abstract classes. Maybe I am wrong in doing so, I don't know how the compiler actually treats them. But still, it totally makes sense to me that this rule handles both.
However, I understand that my point of view may not be shared by everyone. If you think a separate rule should be better, I can open another PR for that instead!

antfu:
Thanks for your contributions. I still think this is out-of-scope of this rule. Feel free to open a proposal to discuss a new rule or so.

Issue #312, opened 2024-04-04
@stylistic/jsx-indent conflicts with @stylistic/indent


yunsii:
### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

### Describe the bug

An error will be reported anyway
![image](https://github.com/antfu/eslint-config/assets/18096089/1d25168c-51fd-47a1-b7ee-e8da3305e381)

### Reproduction

https://github.com/yunsii/stylistic-repro/blob/master/src/index.tsx

### Contributes

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests

PR #368, opened 2024-04-26
fix(jsx-indent): conflict between indent and jsx-indent

Daydreamer-riri:
### Description

Currently, there are some issues with jsx-indent, where some indentations are not checked, and there are conflicts with the indent rules. I have added some tests that cannot be passed.

### Additional context

image
image
image
image

### DIFF
--- a/packages/eslint-plugin-jsx/rules/jsx-indent/jsx-indent.ts
+++ b/packages/eslint-plugin-jsx/rules/jsx-indent/jsx-indent.ts
@@ -380,6 +380,8 @@ export default createRule<MessageIds, RuleOptions>({
       const nameIndent = getNodeIndent(node.name)
       const lastToken = context.sourceCode.getLastToken(node.value)!
       const firstInLine = getFirstNodeInLine(context, lastToken)
+      if (firstInLine.loc.start.line !== lastToken.loc.start.line)
+        return
       const indent = node.name.loc.start.line === firstInLine.loc.start.line ? 0 : nameIndent
       checkNodesIndent(firstInLine as unknown as ASTNode, indent)
     }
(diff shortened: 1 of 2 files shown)


antfu:
The test seems to fail, can you take a look? Thanks

Daydreamer-riri:
Apologies, I realized I overlooked a few things. When I set `style/indent` (which I use with `@antfu/eslint-config`) to `'style/indent': ['error', 2],` most of the issues I raised disappeared. It seems that JSX-related indentation is ignored by default in `@antfu/eslint-config`? I'm curious to know why this is the case.

I only have this one question left at the moment. I'm not sure if this is still due to something I've overlooked.
image

Daydreamer-riri:
Hi @antfu , I would like to know if there is any specific reason why the indent rule defaults to ignore JSX-related nodes. This will determine whether I should implement indentation within the JSXExpressionContainer in the jsx-indent rule.

Daydreamer-riri:
This will fix #312 .

Daydreamer-riri:
This PR is ready.

packages/eslint-plugin-jsx/rules/jsx-indent/jsx-indent.ts
```diff
@@ -380,6 +380,8 @@ export default createRule<MessageIds, RuleOptions>({
       const nameIndent = getNodeIndent(node.name)
       const lastToken = context.sourceCode.getLastToken(node.value)!
       const firstInLine = getFirstNodeInLine(context, lastToken)
+      if (firstInLine.loc.start.line !== lastToken.loc.start.line)
```
antfu, line 383:
```suggestion
      // leave this for `indent` rule to handle
      if (firstInLine.loc.start.line !== lastToken.loc.start.line)
```

antfu (approved):

Issue #378, opened 2024-05-01
`indent`: Support JSXText indent


Daydreamer-riri:
### Clear and concise description of the problem

‌‌‌‌‌‌Based on my practice, the `indent` rule can "almost" replace the `jsx-indent` rule and do a better job. Before I made the fix (#368), there was even a conflict between the two. Moreover, `jsx-indent` would lack indentation in multiline ternary operations. So now I prefer to turn off the `jsx-indent` rule and use only the `indent` rule. However, the reason I used the word "almost" is because `indent` does not handle the following situation:

```jsx
<div>
JSXText  // `indent` does not indent for this
  {'Literal in JSXExpressionContainer'} // `indent` will indent for this node
</div>
```

### Suggested solution

‌‌‌‌‌‌‌Support JSXText indentation for the `indent` rule. Considering that the rule already indents almost all JSX-related nodes (and it handles the overlapping parts of JSX nodes with other nodes quite well), the current lack of support for JSXText might be considered a BUG? I believe this should not be considered a breaking change.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Contributes

- [X] If this feature request is accepted, I am willing to submit a PR to fix this issue

antfu:
Sounds good, feel free to send the PR.

Would you suggest implementing the feature-parity and then deprecating just-indent in favor of indent?

Daydreamer-riri:
Yes, I suggest it. I can attempt to make `indent` pass all the test cases of `jsx-indent` to achieve feature parity.

antfu:
Sounds nice if we managed to do that. Feel free to PR, thanks!

Daydreamer-riri:
The task of implementing jsx-indent testing is more complex than initially anticipated due to the need to consider indentation rules for non-jsx code as well. I am still working on this task as quickly as possible. (I was a bit busy last week🥹)

JstnMcBrd:
Now that this feature was added in 2.2.0, will the `jsx-indent` rule be deprecated as suggested earlier?

[...]

PR #413, opened 2024-06-05
feat(indent): support for JSXText indentation

Daydreamer-riri:
### Description

For specific discussions, see #378.

- Support JSXText indentation for the indent rule.
- ‌‌‌‌‌‌Make the `indent` rule pass all tests of `jsx-indent`.

### Linked Issues

#378


hyoban:
Can we change the shared configurations now? Don't ignore jsx-related nodes and remove `jsx-indent` rule.

https://github.com/eslint-stylistic/eslint-stylistic/blob/c11f1d6f26c13b29fd44c95c908922cb79a0ac82/packages/eslint-plugin/configs/customize.ts#L115-L154

https://github.com/eslint-stylistic/eslint-stylistic/blob/c11f1d6f26c13b29fd44c95c908922cb79a0ac82/packages/eslint-plugin/configs/customize.ts#L238

Daydreamer-riri:
> Can we change the shared configurations now? Don't ignore jsx-related nodes and remove `jsx-indent` rule.

Sure, I will commit as soon as possible.

hyoban:
Or maybe change the configurations in a separate PR?

Daydreamer-riri:
‌‌That's a good idea. I will submit a PR to change the configuration after this PR is merged.

hyoban:
FYI. It seems we can not just remove `JSXOpeningElement` and `JSXClosingElement`, which will produce a wrong indent for `TSTypeParameterInstantiation`.

I need to set different options for js and jsx files. Js files should add them to `ignoredNodes` and jsx files shouldn't.

antfu:
So, does this mean we could deprecate `jsx-indent` then? I think it worth to have a option to opt-out the JSX handing as well

Daydreamer-riri:
> So, does this mean we could deprecate `jsx-indent` then? I think it's worth to have an option to opt-out the JSX handling as well

Currently, `indent` has passed all the `jsx-indent` tests (these tests have been reasonably modified), and I think `jsx-indent` can be deprecated. If there are any issues in the future, I will take responsibility for handling them.

antfu (approved):

PR #418, opened 2024-06-14
Make plugin assignable to ESLint.Plugin

### LINKED ISSUE #398: Types: '@stylistic/eslint-plugin' is not assignable to 'ESLint.Plugin'
### Validations

### Bug

I am trying to create a type-safe ESLint config using `@stylistic/eslint-plugin`. But it appears the types are not formatted correctly.

Here is a simplified version of what I have:
```ts
import stylistic from '@stylistic/eslint-plugin';
import type { Linter } from 'eslint';

export function config(): Linter.FlatConfig[] {
    return [
        {
            plugins: {
                // Throws type error
                // The type of 'stylistic' is not compatible with 'ESLint.Plugin'
                '@stylistic': stylistic,
            },
            // ...
        },
    ];
}
```

Here is the full error message:
```
Type 'typeof import(".../@stylistic/eslint-plugin/dist/dts/index")' is not assignable to type 'Plugin'.
  Types of property 'configs' are incompatible.
    Type '{ 'disable-legacy': FlatConfig<RulesRecord>; customize: { (options: StylisticCustomizeOptions<false>): BaseConfig<RulesRecord, RulesRecord>; (options?: StylisticCustomizeOptions<...> | undefined): FlatConfig<...>; }; ... 4 more ...; 'recommended-legacy': BaseConfig<...>; }' is not assignable to type 'Record<string, FlatConfig<RulesRecord> | FlatConfig<RulesRecord>[] | ConfigData<RulesRecord>>'.
      Property 'customize' is incompatible with index signature.
        Type '{ (options: StylisticCustomizeOptions<false>): BaseConfig<RulesRecord, RulesRecord>; (options?: StylisticCustomizeOptions<true> | undefined): FlatConfig<...>; }' is not assignable to type 'FlatConfig<RulesRecord> | FlatConfig<RulesRecord>[] | ConfigData<RulesRecord>'.
```

I have created a simple reproduction to demonstrate: https://stackblitz.com/edit/github-dbnoms

### Cause

It appears that in [eslint-plugin/configs/index.ts](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/packages/eslint-plugin/configs/index.ts#L21), you have added the `customize()` factory method to the list of configurations. But ESLint's types require that the `ESLint.Plugin["configs"]` property must be of type `Record<string, ConfigData | Linter.FlatConfig | Linter.FlatConfig[]>` - the values must be config objects, not methods.

The type of `plugin.configs` is defined in a separate .dts file - [eslint-plugin/dts/index](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/packages/eslint-plugin/dts/index.ts#L14) - which uses the exported type value of `eslint-plugin/configs/index.ts`, which sends the issue to my type-checker.

### Proposed solutions

It should be possible to define the config's type in such a way that keeps `customize()` but also tells TypeScript that configs is compatible with `ESLint.Plugin["configs"]`.

Locally, I was able to edit the dts files in `node_modules` and defined `config` as an extension of `ESLint.Plugin["configs"]` and the type error went away.

So in [eslint-plugin/dts/index](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/packages/eslint-plugin/dts/index.ts#L14), you could do this:
```diff
- configs: typeof configs
+ configs: ESLint.Plugin["configs"] & typeof configs
```

Of course, you may know of a better solution - I am not familiar with how the TypeScript build process works for this repository. For example, I don't know why there are custom .dts files instead of letting TypeScript generate them. Whatever the reason, perhaps it will influence what solution you choose.

### Reproduction

https://stackblitz.com/edit/github-dbnoms

### Contributes

JstnMcBrd:
### Linked Issues

Closes #398 

### Description

See the linked issue for a description of the problem.

Since the issue hasn't received any attention in a month, I decide to go ahead and try the solution I suggested there.

Again, you may know of a better solution to this problem, or perhaps you don't like this solution. If so, feel free to close this PR and we can try something else instead. I'm not familiar with this repository so I'm eager to accept correction.

### DIFF
--- a/packages/eslint-plugin/dts/index.ts
+++ b/packages/eslint-plugin/dts/index.ts
@@ -1,4 +1,4 @@
-import type { Rule } from 'eslint'
+import type { ESLint, Rule } from 'eslint'
 import type { configs } from '../configs'
 import type { UnprefixedRuleOptions } from './rule-options'
 
@@ -11,7 +11,7 @@ export type Rules = {
 
 declare const plugin: {
   rules: Rules
-  configs: typeof configs
+  configs: ESLint.Plugin["configs"] & typeof configs
 }
 
 export default plugin
--- a/packages/eslint-plugin/dts/index.ts
+++ b/packages/eslint-plugin/dts/index.ts
@@ -11,7 +11,7 @@ export type Rules = {
 
 declare const plugin: {
   rules: Rules
-  configs: ESLint.Plugin["configs"] & typeof configs
+  configs: ESLint.Plugin['configs'] & typeof configs
 }
 
 export default plugin


antfu (approved):

JstnMcBrd:
Whoops, forgot to format the PR title as "fix:", so I guess it won't be included in the automatic changelog [next release](https://github.com/eslint-stylistic/eslint-stylistic/pull/402#issue-2304125024) :(

PR #447, opened 2024-06-25
fix(jsx-indent): deprecate `jsx-indent` rule in favor of `indent`

antfu:
As `indent` rule is now working well with JSX since https://github.com/eslint-stylistic/eslint-stylistic/pull/413, this PR deprecates the `jsx-indent` rule. Usage of it should be replaced with the bear `indent` rule.

/cc @Daydreamer-riri so that you are aware of this


Daydreamer-riri:
‌‌‌‌‌‌‌Great! I will then confirm whether `jsx-indent-props` can also be deprecated.

silverwind:
Are the options `checkAttributes` and `indentLogicalExpressions`from `jsx-indent` also needed or supported on `indent`?

xsjcTony:
> Are the options `checkAttributes` and `indentLogicalExpressions`from `jsx-indent` also needed or supported on `indent`?

Same question. Nothing is mentioned in changelog / doc

Daydreamer-riri:
Hi @silverwind @xsjcTony , the `checkAttributes`, `indentLogicalExpressions` options have the same functionality when the `indent` rule does not ignore related nodes. To not check for related nodes, set the `indent` rule's `ignoredNodes` option.
For example if you don't want to check jsxAttributes nodes, you can add `JSXAttribute` to `ignoredNodes`.

If there are still any problems, please let me know and I will reply as soon as possible!

silverwind:
Okay, so `checkAttributes: true` is equivalent to `JSXAttribute` being absent in `ignoredNodes`, but I still don't quite follow what to do about `indentLogicalExpressions: true`, is there an equivalent option?

Daydreamer-riri:
```js
{
  ignoredNodes: [
    // checkAttributes
    'JSXAttribute',
    // indentLogicalExpressions
    'LogicalExpression[right.type="JSXElement"] > JSXElement' > 'JSXOpeningElement',
    'LogicalExpression[right.type="JSXFragment"] > JSXFragment' > 'JSXOpeningFragment',
  ]
}
```

silverwind:
Thanks, so it looks like the default options of `indent` are already resulting in the previous behaviour with `{checkAttributes: true, indentLogicalExpressions: true}`, so in my case, there was nothing to change. Above example is for people who want the previous defaults of `jsx/indent` where both options were false.

PR #465, opened 2024-07-14
fix(indent): handle mixed spaces and tabs

hyoban:
### Description

As far as I know, `no-mixed-spaces-and-tabs` does not provide auto-fix, I don't know when they would conflict. At the same time, this prevents the `indent` rule from fixing errors when tabs and spaces are mixed.

I tested it in my ESLint config, it works fine.

https://github.com/hyoban/eslint-config-hyoban/blob/5b6a5301b9c1db0f2daa8605e4c30f92b20f4abb/patches/%40stylistic__eslint-plugin-js%402.6.0-beta.0.patch

### DIFF
--- a/packages/eslint-plugin-js/rules/indent/indent.test.ts
+++ b/packages/eslint-plugin-js/rules/indent/indent.test.ts
@@ -2102,30 +2102,12 @@ run({
     {
       code: $`
         function foo() {
-          bar();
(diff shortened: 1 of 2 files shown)


antfu:
Sounds good. But we might need to wait until the next minor to have this

antfu (approved):

PR #466, opened 2024-07-15
feat!: migrate js/ts/jsx to single package

9romise:
### Description

This PR comes from https://eslint.style/contribute/project-progress#%F0%9F%91%89-4-maintenance-refactors for fusing JS/TS/JSX rules into a single package

This PR is based on `dev` branch to allow a smooth merge after the upgrade of typescript-eslint@8

- [ ] move ts rules to plugin-js
- [ ] migrate js rules' `createRules` to `ESLintUtils.RuleCreator` in `@typescript-eslint/utils` (?)
- [ ] move js and jsx rules into `eslint-plugin/rules` and make it a single package (?)
- [ ] sorted out user-facing configurations, compatibility, etc
- [ ] update docs to introduce this breaking change


packages/eslint-plugin-js/rules/comma-spacing/ts-comma-spacing.test.ts
```diff
@@ -410,19 +410,6 @@ run({
         },
       ],
     },
-    {
```
9romise, line 413: Conflicts with valid test case in the js version
https://github.com/eslint-stylistic/eslint-stylistic/blob/240006cd887de265dc33eb027765846bd877bfd2/packages/eslint-plugin-js/rules/comma-spacing/comma-spacing.test.ts#L89

9romise:
There are some differences in optional call in the `function-call-spacing` of js and ts version. We have to choose a reasonable result and modify the test case. cc @antfu 
image
image

antfu:
Thanks for your work on it! Tho, I am afraid it might be a bit dangerous to do so many changes at once. Also, consider this refactor will be working in parallel for a long time with other PRs coming in at the same time.

I would suggest we take multiple stages to do this:

## Stage 1

We can first colocate js and its implementations inside the same package, but keep them as separate files. For example:

```
|- packages/eslint-plugin/rules
   |- indent
      | indent-js.ts
      | indent-ts.ts
      | indent.ts
      | indent-js.test.ts
      | indent-ts.test.ts
```

Where we also have `package/eslint-plugin-js/indent/indent.ts` re-export from `package/eslint-plugin/indent/indent-js.ts`

We don't touch the content of those files at all, but only the file structure. This way we can move forward and doing releases without changing the runtime implementation at all.

## Stage 2

We can then refactor those files under each folder and make sure the ts version passes both js/ts tests. We can start to use the `indent-ts` to replace the `js` rules

## Stage 3

We merge the implantation of `indent-js` and `indent-ts` without touching the tests, where the tests guard us to make sure we don't break the behavior.

----

Each stage could come with multiple smaller PRs, where we could merge and move forward. Between stages, we could do minor or major to make sure the changes are well communicated to the users.

What do you think?

9romise:
I absolutely agree with you, I admit that this PR is indeed a bit too tough.

Here are some of my other thoughts:
1. Stage 1 should be completed as early as possible to avoid conflicts with other PRs modifying the rules. Handling these conflicts before the release might take some time. Perhaps you can complete it first, but if you don't have the time, I would be happy to create a PR.
2. Based on the work of this PR, I found that some rules have inconsistent performance between JS and TS versions. In Stage 2, we can merge the consistent rules first, and then deal with the inconsistent rules separately. By that time, we might need some discussions to decide on the final behavior of these rules.

PR #477, opened 2024-07-24
Allow to configure "padding-line-between-statements" rule with the "yield" operator

phansys:
### Description

Allow to configure "padding-line-between-statements" rule with the "yield" operator.
By instance:

```js
'padding-line-between-statements': [
  'error',
  {
    blankLine: 'always',
    prev: '*',
    next: ['break', 'continue', 'return', 'throw', 'yield']
  }
]
```

### To Do
- [ ] Add tests.


phansys:
Hi :wave: 
As I'm not sure if the proposed implementation is acceptable or it is pointing to the right direction, I'd love to receive some feedback before going further.

Thank you!

PR #499, opened 2024-08-09
fix(arrow-parens): wrong parens removal with optional parameter

### LINKED ISSUE #498: arrow-parens : wrong parens removal with optional parameter
### Validations

### Describe the bug

myObj.something = (opts?) => doSomething(opts);
Tries to remove the parens which makes the code syntax incorrect

### Reproduction

https://stackblitz.com/edit/github-fpztgp?file=src%2Findex.ts

### Contributes

9romise:
### Linked Issues

Closes #498 

### Additional context

Should we continue to test this way?
It uses some immutable files as parsing results, which cannot be verified with the latest version of parser.
At the same time, this requires us to keep these files in the repo. I think this is unnecessary.


antfu (approved):

antfu:
Yeah I would love to refactor them away and use real parsers for sure.

PR #519, opened 2024-08-29
refactor: make ts plugin self-contains

antfu:


antfu:
Make the ts plugin directly bundle js rules instead of requiring the entire package.

Issue #482, opened 2024-07-29
Merge js/ts/jsx into single package


antfu:
https://github.com/eslint-stylistic/eslint-stylistic/pull/466#issuecomment-2254248750

## Stage 1

We can first colocate js and its implementations inside the same package, but keep them as separate files. For example:

```
|- packages/eslint-plugin/rules
   |- indent
      | indent-js.ts
      | indent-ts.ts
      | indent.ts
      | indent-js.test.ts
      | indent-ts.test.ts
```

Where we also have `package/eslint-plugin-js/indent/indent.ts` re-export from `package/eslint-plugin/indent/indent-js.ts`

We don't touch the content of those files at all, but only the file structure. This way we can move forward and doing releases without changing the runtime implementation at all.

## Stage 2

We can then refactor those files under each folder and make sure the ts version passes both js/ts tests. We can start to use the `indent-ts` to replace the `js` rules

## Stage 3

We merge the implantation of `indent-js` and `indent-ts` without touching the tests, where the tests guard us to make sure we don't break the behavior.

9romise:
It's exciting to see that all the rules have been merged into one version now! We can start deprecating sub-packages and refactoring the project structure.

darksabrefr:
Hum... since https://github.com/eslint-stylistic/eslint-stylistic/pull/793, we now have a deprecation message showing your intentions. But are you aware that some people using `@stylistic/eslint-plugin-js` instead the global `@stylistic/eslint-plugin` because it has far less sub-dependencies, and in particular no Typescript related ones ? Moreover it imposes to add `typescript` in the project direct dependencies as this one is a peer dependency of `@typescript-eslint/utils`, one the exclusive dependencies of `@stylistic/eslint-plugin`.
There's still pure JS projects today and these changes are not very happy for them. The https://github.com/eslint-stylistic/eslint-stylistic/issues/571 issue describing this problem has not yet been resolved

antfu:
Let's keep #571 open, and feel free to contribute if you have better solutions. The decision of to merge is made at the moment this project is created. We are doing it for better maintenance in the long run and also for simplified usage. TypeScript is gradually becoming the de facto standard. It might not be the most ideal situation where `typescript` is introduced as a dep, but I also not sure if the cost of solving that is worth it as the benefit is rather marginal. That said, we are probably counting on the community to send PR to improve that as it's not our main focus.

PR #529, opened 2024-09-01
test: cross test merged rules (rest) 

9romise:
Part of https://github.com/eslint-stylistic/eslint-stylistic/issues/482 Stage 2

This PR fixes the failing rules marked in https://github.com/eslint-stylistic/eslint-stylistic/pull/520.


packages/eslint-plugin/rules/padding-line-between-statements/padding-line-between-statements._js_.test.ts
```diff
@@ -3,8 +3,7 @@
  * @author Toru Nagashima
  */
 
-// TODO: Stage 2: Test merged rule
-import rule from './padding-line-between-statements._js_'
```
antfu, line 7: This rule is kinda special as the TS version does not inherit the JS at all. That cause the coverage to fail - which I think it's fine, but maybe at a comment in the JS file to remind us to remove that file in the future

antfu:
We can also do rule-by-rule for each PR, so we go progressively

9romise:
Now there are three rules left that are difficult to solve, I leave some comments. We can fix them one by one.

PR #534, opened 2024-09-08
docs: typo

yoavbls:
### Description

Unnecessary comma cause diff addition not to show + missing @ before `stylistic`:  
  
CleanShot 2024-09-09 at 02 03 07@2x

### DIFF
--- a/docs/guide/migration.md
+++ b/docs/guide/migration.md
@@ -268,7 +268,7 @@ export default [
       stylistic: StylisticPlugin
     },
     rules: {
-      'stylistic/semi': 'error',
+      '@stylistic/semi': 'error',
       // ...
     }
   }
--- a/docs/guide/migration.md
+++ b/docs/guide/migration.md
@@ -268,7 +268,7 @@ export default [
       stylistic: StylisticPlugin
     },
     rules: {
-      '@stylistic/semi': 'error',
+      'stylistic/semi': 'error',
       // ...
     }
   }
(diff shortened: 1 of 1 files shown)


docs/guide/migration.md
```diff
@@ -261,14 +261,14 @@ export default [
...
-  StylisticPlugin.configs['disable-legacy'] // [!code ++],
+  StylisticPlugin.configs['disable-legacy'] // [!code ++]
   // your own rules
   {
     plugins: {
       stylistic: StylisticPlugin
     },
     rules: {
-      'stylistic/semi': 'error',
+      '@stylistic/semi': 'error',
```
antfu, line 271: No, this is not how it works. The previous example is correct as we provide the plugin named as `stylistic`

docs/guide/migration.md
```diff
@@ -261,14 +261,14 @@ export default [
     ],
   }),
   // override the legacy rules
-  StylisticPlugin.configs['disable-legacy'] // [!code ++],
+  StylisticPlugin.configs['disable-legacy'] // [!code ++]
```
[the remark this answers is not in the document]
yoavbls: fixed

docs/guide/migration.md
antfu: `@stylistic` was mainly the convention of the legacy config, where in the flat config I don't want users to carry that legacy convention when they don't need to.

PR #544, opened 2024-09-23
fix(quotes): quotes options execute `allowTemplateLiterals` first then `avoidEscape`

### LINKED ISSUE #542: `quotes`: allowTemplateLiterals false does not work properly
### Validations

#### eslint.config.mjs

```js
import ESLintPluginStylistic from "@stylistic/eslint-plugin";

export default {
  plugins: {
    "@stylistic": ESLintPluginStylistic,
  },

  rules: {
    "@stylistic/quotes": [
      "warn",
      "double",
      {
        avoidEscape: true,
        allowTemplateLiterals: false,
      },
    ],
  },
};
```

#### main.js

```
const a = `"a"`;
```

- The rule does not warn on this line, which is a bug.
- It should auto-fix it to: `const a = '"a"'`
- I think this is a regression, because this case seemed to be correctly handled when I used `@typescript-eslint/quotes` (before they deprecated it).

### Contributes

- [X] I am willing to submit a PR with failing tests

zzxming:
### Description

only when `allowTemplateLiterals` set true, allow template literals to avoid escape sequences 

### Linked Issues

close #542

### DIFF
--- a/packages/eslint-plugin/rules/quotes/quotes._js_.test.ts
+++ b/packages/eslint-plugin/rules/quotes/quotes._js_.test.ts
@@ -21,8 +21,8 @@ run({
     { code: 'var foo = "bar";', options: ['single', { ignoreStringLiterals: true }] },
     { code: 'var foo = "\'";', options: ['single', { avoidEscape: true }] },
     { code: 'var foo = \'"\';', options: ['double', { avoidEscape: true }] },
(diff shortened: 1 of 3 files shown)


antfu (approved):

Zamiell:
Thank you zzxming!

PR #548, opened 2024-09-25
feat(curly-newline): add new rule

### LINKED ISSUE #540: `padded-blocks`: add line break after block open curtly and before close curly brackets

Eslint still unable to format blocks like this:
```
{ const a = 1;
      a ++;}
```
This block leaves unchanged after formatting.

Expected result is:
```
{
    let a = 1;
    a++;
}
```

### Suggested solution

Is it possible to improve `padded-blocks` or create new rule which will just add missing line breaks for such blocks?
`padded-blocks` currently can add new lines, that is not always needed.

### Alternative

Currently I need to run prettier before eslint to format such blocks properly.

### Additional context

_No response_

### Validations

### Contributes

keijokapp:
### Description

Here's my take at solving #540 and the problems I have with current solutions for other types of blocks (ie. `brace-style`). This rule is modeled after [`object-curly-newline`](https://eslint.style/rules/default/object-curly-newline). It handles all kinds of blocks rather than just the lone ones. It offers a way to specify different options for each of the 23 block types and common patterns I found. For me, the ability to allow single-line blocks in some cases is necessary but the alternatives (eg `brace-style` with `allowSingleLineBlock: false`) do not allow that.

The rule name follows the pattern in ESLint that "curly" refers to curly braces enclosing block-like structures (only example is [`curly`](https://eslint.org/docs/latest/rules/curly) though) and "object-curly" refers to curly braces enclosing object-like structures. ESLint rules generally do not include "block" in their name.

Ideally, `brace-style` and `padded-blocks` would ignore single-line blocks and leave them up to this rule (at least by default). Maybe it's something to consider when the rules eventually start diverging from their deprecated ESLint counterparts. [ESLint's semantic versioning policy](https://github.com/eslint/eslint?tab=readme-ov-file#semantic-versioning-policy) allows doing such changes in a minor release.

### Linked Issues

Resolves #540

### Additional context


keijokapp:
The rule conflicts some other rules as defined in `all` config. That's why the fixture test is failing. I don't have a good solution here. Tweaking the defaults of this rule just to work with other rules' defaults doesn't seem a good solution.

[The documentation](https://eslint.style/guide/config-presets#enable-all-available-rules) already acknowledges the problem with `all` config. The rules have been developed over a long period of time by different people with different philosophies.  I'm wondering about a couple of things:
1. Is there a reason to provide this config at all? Is it reasonable to add new rules to that config? Is it reasonable for this preset to only use default/undefined options?
2. Wouldn't it make sense for `eslint-plugin` have it's own version of some rules (decoupled from JS/TS/JSX variants)? That way, the behavior of the original rules can be frozen for those wanting a simple migration from ESLint/`typescript-eslint` while `eslint-plugin` can move in its own direction. As long as semantic versioning is followed, the defaults and behavior of the rules could be changed to make the rules work together.

antfu:
That's a very thoughtful PR, thanks a lot!

For the `all` config - Honestly, I would love to deprecate it at all. This was an old tradition in the ESLint ecosystem, as users might try to get all the "sensible defaults" from the presets without too many configurations. However, as time goes on and the language evolves, that approach doesn't always work. Before that, we could have some special conditions, maybe to say we exclude this rule from the `all` config.

There are some confusion around the rule's default (inherited from ESLint and ts-eslint), and the defaults we recommend (with `customize()`). While I would love to change the default to make them more aligned - I doubt if that would worth it from a community perspective - as it would basically force everybody's config to change with no actually benefits. In that case I don't think we could ever do that - unless we could find a good way and probably a codemod to help migrate smoothly (but again that would still be too much work and risk)

antfu (approved):

Issue #211, opened 2023-12-07
`max-statements-per-line`: allow specific statements


Cherry:
### Clear and concise description of the problem

It would great if this rule could ignore specific statements, such as `continue` and `break`, since these are often used in loops and it's a common stylistic choice from my experience to do these on a single line.

For example, a contrived loop. There's much better ways to do this in modern JS, but illustrates the request:
```javascript

const items = [
	{
		include: true,
		value: 1,
	},
	{
		include: false,
		value: 2,
	},
	{
		include: true,
		value: 3,
	},
];

for (const item of items) {
	if (!item.include) { continue; } // errors with @stylistic/max-statements-per-line: ["error", { "max": 1 }]
	console.log(item.value);
	// do something with item
}

```

While I do want to prevent things like:
```javascript

if (condition) { bar = 1; } else { bar = 2; }

// or

for (var i = 0; i < length; ++i) { bar = 1; }
```

I think allowing `break` or `continue` as exceptions to this could be worthwhile

### Suggested solution

Perhaps something like this could be a useful configuration?

```js
'@stylistic/max-statements-per-line': [
		'error',
		{
			max: 1,
			ignoredStatements: [
				'ContinueStatement',
				'BreakStatement',
			],
		},
	],
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Contributes

- [ ] If this feature request is accepted, I am willing to submit a PR to fix this issue

Cherry:
Hmm yeah that definitely feels odd. I wouldn't have expected the wrapping in braces to produce different results 🤔

The primary difference in the AST appears to be that the first has the `ContinueStatement` as the `consequent`, whereas the second has a `BlockStatement` with the `ContinueStatement` in the `body`.
![](https://up.jross.me/byw9owcl)
https://astexplorer.net/#/gist/eac93feb074c59397c23955caa182f84/b62518fdd7cbc027c195b2496617b3ea0e068b16

500-internal-server-error:
[...]

PR #558, opened 2024-10-10
feat(max-statements-per-line): new option to allow specific statements

zzxming:
### Description

add a new option to ignore specific statement code check

### Linked Issues

#211

### DIFF
--- a/packages/eslint-plugin/rules/max-statements-per-line/max-statements-per-line._js_.ts
+++ b/packages/eslint-plugin/rules/max-statements-per-line/max-statements-per-line._js_.ts
@@ -27,7 +27,7 @@ export default createRule<RuleOptions, MessageIds>({
             minimum: 1,
             default: 1,
           },
-          ignoredNodes: {
+          listeningNodes: {
             type: 'array',
             items: {
               type: 'string',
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/rules/max-statements-per-line/max-statements-per-line._js_.ts
```diff
@@ -174,5 +181,13 @@ export default createRule<RuleOptions, MessageIds>({
       'ExportAllDeclaration:exit': leaveStatement,
       'Program:exit': reportFirstExtraStatementAndClear,
     }
+
+    for (const key in listeners) {
+      if (ignoredNodes.includes(key) || ignoredNodes.includes(`${key}:exit`)) {
+        delete listeners[key]
```
antfu, line 187: Maybe we should have a string array to list all the enter statements, and list them as enum in the config, and then construct the listener instead of removing them. For better UX and more robustness.

antfu (changes requested):

packages/eslint-plugin/rules/max-statements-per-line/max-statements-per-line._js_.ts
zzxming: I'm not sure if I understand what you mean, is it like I changed it?

antfu: Yes, the direction is what I was imagining. But I think instead of of `listeningNodes` I think it would be easier to use `ignoredNodes` to exclude

antfu (approved):

packages/eslint-plugin/rules/max-statements-per-line/max-statements-per-line._js_.ts
zzxming: oh, that's what you mean. this is indeed much better. thank you for your help

PR #565, opened 2024-10-21
feat!: cross test `function-call-spacing`

9romise:
### Description

> BREAKING CHANGE IN **TS** VERSION
> SYNC WITH **JS** VERSION

Auto-fixing optional chain is different in the two versions.

Changes:
- update report location
- don't remove comments while auto-fixing (#588)
- support `ImportExpression` in `ts` version (#570)
- auto-fix optional chain when the option is `never` (#589)
- add `optionChain` to control `f?. ()`, `f ?.()`, `f ?. ()` when the option is `always` (#605)
- auto-fix while there is a line break and the option is never (#606)

### Linked Issues

part of #482


antfu:
Maybe we could make a factory function to create two version of rules with different defaults, use it to create two different entry?

9romise:
After careful consideration, I think that this PR contains too much content. Maybe I should split it up and submit.
Here's what I think about the changes.

> - update report location
> - don't remove comments while auto-fixing
> - support ImportExpression in ts version
> - auto-fix optional chain when the option is never

Some safe changes (or maybe fixes)


> - allow f?. (), f ?.(), f ?. () when the option is always

Since the core problem is the difference in the optional chain, I think maybe we can just add an option for the user to choose, which can also get a better auto-fix.
Example: `@stylistic/function-call-spacing: ["error", "always", { "optionalChain": { "before": true, "after": true } }]`


> - don't auto-fix while there is a line break and the allowNewlines option is falsy

There are some syntax errors in the issue description, so we can do without porting it.
https://github.com/eslint/eslint/issues/7787

Completing these changes one by one may reduce the breaking changes.

antfu:
Oh yeah that would do. We could merge those non-breaking PRs and ship them, while we could mark all the needed breaking PRs draft (and keep them as small as possible), and finally merge then when we do majors.

antfu (approved):

Issue #545, opened 2024-09-24
`indent-binary-ops`: Indentation Adjustment When ‘(’ Is Included in the Logical Expression


k35o:
### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

### Describe the bug

There was a formatting behavior related to indentation that changed due to modification A, which caught my attention https://github.com/eslint-stylistic/eslint-stylistic/issues/530, which caught my attention.

after
```ts
const a = (
  (flag1
    && flag2)
    || (flag3
      && flag4)
)
```
before
```ts
const a = (
  (flag1
    && flag2)
  || (flag3
    && flag4)
)
```
Indentation is now deeper when logical expressions are at the same level but contain `(`.

I would like each logical expression to be indented at the same level when using comparison operators., even if it contain `(`.
```ts
// Cases without (
const a = (
  flag1
  || flag2
  || flag3
)

//  Hope it's the same as above.
const b = (
  (flag1
    && flag2)
  || (flag3
    && flag4)
  || (flag5
    && flag6)
)
```

I think this is because `needAdditionIndent` is determined to be `true` in [this section](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/packages/eslint-plugin/rules/indent-binary-ops/indent-binary-ops._plus_.ts#L84).
```
// Before the left token is a opening bracket
(['[', '(', '=>', ':'].includes(tokenBeforeAll?.value || '') && firstTokenOfLineLeft?.loc.start.line === tokenBeforeAll?.loc.start.line)
```

### Reproduction

https://stackblitz.com/edit/github-d1geg2?file=src%2Findex.ts

### Contributes

- [X] I am willing to submit a PR to fix this issue
- [X] I am willing to submit a PR with failing tests

Anoesj:
Not entirely sure this is the same or a similar scenario, but this one caught my attention when trying to configure Stylistic to taste:
![image](https://github.com/user-attachments/assets/1a9f0634-b5c0-44f0-8ca1-5ef95d4355e6)

I use the following settings for `indent`:
```
'@stylistic/indent': ['error', 2, {
  VariableDeclarator: 'first',
  SwitchCase: 1,
}],
```

[...]

PR #576, opened 2024-10-28
feat(indent-binary-ops): logical expression indent

k35o:
fixed https://github.com/eslint-stylistic/eslint-stylistic/issues/545

### Description
I want each logical expression to be indented at the same level when using comparison operators

Current specification does not allow one conditional expression to make the indentation shallower than the previous line.
The indentation is either maintained or deepened by one.
This is troubling because even comparisons at the same level are indented at different levels

This PR changes this so that comparisons at the same level start at the same indentation.

### Linked Issues
https://github.com/eslint-stylistic/eslint-stylistic/issues/545

### Additional context
Logic to reduce indentation below the previous line has been added.
The condition for lowering the indentation is that the previous expression is on more than one line
The previous expression being multiple lines is determined by the previous token ending in `)` and the parenthesis pair not being completed on the previous line.

I feel there is a better and more accurate algorithm, please let me know if there is one.

For example, the following cases do not adjust indentation well.
```ts
// input
a = (
  (b
      && c)
  [
    0, 1, 2]
    || (d
  && e)

// output
const a = (
  (b
    && c)
  [
    0, 1, 2]
    || (d
      && e)
)

// wanted
"const a = (
  (b
    && c)
  [
    0, 1, 2]
  || (d
    && e)
)"
```

We consider it difficult to make another case work if `]` is targeted in addition to `)`.


k35o:
There are many areas where the formatting is working poorly and will be corrected.

k35o:
I fixed it.

k35o:
@antfu 
Can you rerun the test?

PR #590, opened 2024-10-30
fix: correct the type of `quotes` in `customize`

### LINKED ISSUE #568: `stylistic.configs.customize({ quotes: "backtick" })` gives type error even though it's valid
### Validations

### Describe the bug

`eslint.config.js`
```js
import stylistic from "@stylistic/eslint-plugin"

export default [ stylistic.configs.customize({ quotes: "backtick" }) ]
```
reports
```
No overload matches this call.
  Overload 1 of 2, '(options: StylisticCustomizeOptions<false>): BaseConfig<RulesRecord, RulesRecord>', gave the following error.
    Type '"backtick"' is not assignable to type '"single" | "double" | undefined'.
  Overload 2 of 2, '(options?: StylisticCustomizeOptions<true> | undefined): Config<RulesRecord>', gave the following error.
    Type '"backtick"' is not assignable to type '"single" | "double" | undefined'.ts(2769)
```
even though it's a valid option

### Reproduction

to reproduce, use the config from above

### Contributes

9romise:
### Linked Issues
close #568

### DIFF
--- a/packages/eslint-plugin/dts/options.d.ts
+++ b/packages/eslint-plugin/dts/options.d.ts
@@ -22,7 +22,7 @@ export interface StylisticCustomizeOptions<Flat extends boolean = true> {
    *
    * @default 'single'
    */
-  quotes?: 'single' | 'double'
+  quotes?: 'single' | 'double' | 'backtick'
   /**
    * Whether to enable semicolons
    * Similar to `semi` option in Prettier


packages/eslint-plugin/dts/options.d.ts
```diff
@@ -22,7 +22,7 @@ export interface StylisticCustomizeOptions<Flat extends boolean = true> {
    *
    * @default 'single'
    */
-  quotes?: 'single' | 'double'
+  quotes?: 'single' | 'double' | 'backtick'
```
9romise, line 25: Should we use `QuotesSchema0` directly?

antfu: Maybe not, the point of this factory is to hide the implementation details of rules

antfu (approved):

Issue #549, opened 2024-09-26
`jsx-sort-props`: autofix doesn't work if there is comment between props


rakleed:
### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

### Describe the bug

Autofix does not work if there is a comment between props.

![CleanShot 2024-09-27 at 01 22 13](https://github.com/user-attachments/assets/11252058-bd90-440f-ad02-7336c87c2d0e)

```sh
npm run install && npm run lint:fix
```

### Reproduction

Test case:
```js
    {
      code: `
        <Button
          label="count is"
          count={count}
          // color="red"
          setCount={setCount}
        />
      `,
      errors: [expectedError],
      output: `
        <Button
          count={count}
          label="count is"
          // color="red"
          setCount={setCount}
        />
      `,
    },
```
### Contributes

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests

9romise:
Do you have any good ideas on how to fix it, not just the special case in the image above?

9romise:
`jsx-sort-props` has been deprecated. We recommend using the `sort-jsx-props` rule from the `eslint-plugin-perfectionist` package instead. Therefore, I will close this issue.

PR #619, opened 2024-11-25
test(jsx-sort-props): add test with comments between props

rakleed:
### Description

Add test for `jsx-sort-props` with comments between props

### Linked Issues

#549


rakleed:
@antfu could you please approve workflows?

PR #645, opened 2024-12-21
refactor(indent-binary-ops)!: remove unnecessary logic

9romise:
### Description
~Add support for logical operator in es2020~

I've  just found that the logic of this line (contributed by me) is wierd, and removing it has no effect on the test cases, maybe it's unnecessary, but we need more test cases to validate.


tim-wovn:
Hello. Thank you for your work on this PR.

I saw in the 3.0 release that this was labeled as a breaking change. What is the breaking change?

9romise:
> @tim-wovn this change marked an issue in our codebase which I'm not actually sure was the intention? I don't know if this is why it was labeled a breaking change or if this is just a missed edge case in this pr. This is what happens when running `--fix` on the issue it is raising for us.
> 
> Screenshot 2025-02-18 at 11 31 59
> 

Would you like to open a new issue for this case? I'm willing to submit a PR to fix it if somthing wrong.

maddieholtzer:
@9romise Will do. Need to try and get a reproduction in `eslint-stylistic` of what I'm seeing in our code base first.

tim-wovn:
> I'm very sorry, in fact it was mislabeled for breaking changes. It just removes code that doesn't affect the test case and I think the logic is weird.

Thank you for clarifying! 🙇 

At my work we upgraded a few weeks ago and haven't noticed anything unusual. Thank you!

PR #667, opened 2025-01-19
feat(type-generic-spacing): remove whitespaces in generic types

### LINKED ISSUE #244: Space in generics - similar to `space-in-parens`

The following code throws error for the space between `(` and `param1` when `space-in-parens` enabled. But there's no rule that catches the space between `<` and `F,` in the generic scope
```ts
function generic2< F, T> ( param1: F, param2: T) {}
```

### Suggested solution

Having a new rule or adding a support for angle/generic brackets to `space-in-parens` rule. I'm not sure about adding a support for parens rule since the name can be confusing. 
```ts
function generic2<F, T> (param1: F, param2: T) {}
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

### Contributes

### LINKED ISSUE #480: `ts/space-before-function-paren` / `ts/space-before-function-generic`: Add a rule to have spaces before a generic's `<`

I'm looking to have my code formatted as follows
```typescript
function bark <Animal> (animal: Animal) {}
```

But, with the rules available today I only seem to be able to get the following which I believe looks very imbalanced
```typescript
function bark<Animal> (animal: Animal) {}
```

### Suggested solution

Either in rule `ts/space-before-function-paren` an option for `generic` could be provided, or as a separate rule like `ts/space-before-function-generic` which would be similar to `ts/space-before-function-paren` but working with the generic's opening `<`

### Alternative

_No response_

### Additional context

_No response_

### Validations

### Contributes

### LINKED ISSUE #649: `space-in-generics`: add a rule to remove extra spaces in TypeScript generics

The proposal is to add a rule that will remove spaces in TypeScript generics.

Examples:

```ts
interface Foo< T > {} // Bad
interface Foo<T> {} // Good

type Bar< T, U > = T & U; // Bad
type Bar<T, U> = T & U; // Good

function foo< T >(arg: T): T { return arg; } // Bad
function foo<T>(arg: T): T { return arg; } // Good


const x: Array < Array < string > > = []; // Bad
const x: Array<Array<string>> = []; // Good
```

### Suggested solution

Similar rule:

[@stylistic/space-in-parens](https://eslint.style/rules/default/space-in-parens)

My implementation:

[Rule](https://github.com/vnva/eslint-plugin/blob/main/src/rules/space-in-generics.ts)
[Tests](https://github.com/vnva/eslint-plugin/blob/main/src/rules/space-in-generics.test.ts)

### Validations

### Contributes

Ericlm:
### Description

This PR partially solves some common issues with spacing and generic types in TS code. The idea is to fix spacing between `<` and `>` tokens **within** the type declaration (trying to fix it outside is another subject I think, as I could conflict with other spacing rules).

Today, the following syntaxes can be found without any issue : 
```ts
<T>
< T >
<T >
<     T  >
```
This PR tries to fix all of this cases by only allowing `<T>` syntax when using `type-generic-spacing` rule.

I know this may be an opiniated choice as some may prefer `< T >` for example, but for now, I didn't add any option (like the original rule).

### Linked Issues

fix #480 
fix #244 
fix #649 

### Additional context
There are multiple rules that can be impacted by adding spaces around generics (like function definition), so I only focused on spaces within the generic type definition.
I modified the `type-generic-spacing` rule instead of other ones (like `type-annotation-spacing` or creating a new rule), because the added behaviors only affect spacing in generic types.


antfu (approved):

antfu (approved):

swantzter:
I'm not sure this solves #480 given the description of the PR, that issue is about specing outside the `<` `>` which this PR claims to not solve. But I haven't yet had a chance to test

PR #673, opened 2025-01-29
fix(deps): unpin typescript-eslint deps

### LINKED ISSUE #671: `@typescript-eslint/*` deps have been pinned to 8.13.0 in v3.0.0
### Validations

### Describe the bug

`@typescript-eslint/parser` and `@typescript-eslint/utils` deps have been pinned to 8.13.0
https://github.com/eslint-stylistic/eslint-stylistic/commit/89e075a3de9008d2c24ee988d8eac45125a352a1

where previously it allowed open-end range of ^8.13.0.

The change affects the recently released [v3.0.0](https://github.com/eslint-stylistic/eslint-stylistic/releases/tag/v3.0.0).

Was it a mistake or are the deps pinned on purpose? Is there a bug on the typescript-eslint side that affect any of stylistic rules?

### Reproduction

install `@stylistic/eslint-plugin@3.0.0`, look into lockfile

### Contributes

falkenhawk:
### Description

This is to unpin `@typescript-eslint` deps back to open-end range, like it was before [v3.0.0](https://github.com/eslint-stylistic/eslint-stylistic/releases/tag/v3.0.0) - to avoid installing two different versions of typescript-eslint in projects using @eslint-stylistic. Even in the pnpm-lock file there were two versions specified next to each other - 8.13.0 and 8.21.0.

I don't know why those dependencies have been pinned in https://github.com/eslint-stylistic/eslint-stylistic/commit/89e075a3de9008d2c24ee988d8eac45125a352a1#diff-18ae0a0fab29a7db7aded913fd05f30a2c8f6c104fadae86c9d217091709794cR32-R33

if there is a reason, please share your findings.

### Linked Issues
fixes #671

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,8 @@
     "vue-tsc": "catalog:"
   },
   "resolutions": {
-    "@typescript-eslint/utils": "catalog:",
+    "@typescript-eslint/parser": "8.13.0",
+    "@typescript-eslint/utils": "8.13.0",
     "rollup": "catalog:",
     "twoslash-eslint": "catalog:",
     "unbuild": "catalog:",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -88,8 +88,8 @@ catalogs:
       specifier: ^7.5.8
       version: 7.5.8
     '@typescript-eslint/parser':
-      specifier: 8.13.0
-      version: 8.13.0
+      specifier: ^8.13.0
+      version: 8.21.0
     '@unocss/reset':
       specifier: ^65.4.3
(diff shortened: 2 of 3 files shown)


antfu:
Ah, it was because a bug was introduced in the newer version, which caused our tests to fail, I am still waiting for the fix to be landed: https://github.com/typescript-eslint/typescript-eslint/pull/10649 (which should only affect the tests but not on end-user side, which should be safe for people to use the new version)

I was doing that for pinning that for testing but forgot to unpin, thanks for catching it up. I pushed a commit to change how we pinning the version

PR #689, opened 2025-02-18
refactor: update the rules' entry file

9romise:
### Description

Use the ts version rule in the js package without touching the test file.
Then we can merge the rules into a single file one by one in `eslint-plugin`.

### Linked Issues

Part of #482


antfu:
This would make `eslint-plugin-js` `eslint-plugin-ts` `eslint-plugin` effectively the same? I am worried about if this would introduce behaviour breaking changes to the js plugin. I feel we should keep the js plugin as-is, but porting the logics to the ts version (we could add a comment mark in the js version to indicate it's already merged, where we could remove them later when we deprecate the `eslint-plugin-js` plugin

antfu:
Oh do you mean to rename the file that only has the js version? If so I think it's a nice migrate strategy.

packages/eslint-plugin-js/rules/block-spacing/block-spacing.ts
```diff
@@ -1 +1 @@
-export { default } from '../../../eslint-plugin/rules/block-spacing/block-spacing._js_'
```
antfu, line 1: For example, for this rule, I think we shouldn't do this yet.

9romise:
Uh, yes, I do want to remove the suffix from the rule where there is only one version, and this is just the first step.
At the time, I was thinking that the ts version passed all the js test cases, so it wouldn't have a big impact, but maybe it would be better to keep it in its current state for the time being.

Issue #661, opened 2025-01-14
`@stylistic/eslint-plugin` is CJS-only


Kamillaova:
### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

### Describe the bug

Currently, the `@stylistic/eslint-plugin` package is CJS-only. That's bad in 2025, and also I saw this: https://github.com/eslint-stylistic/eslint-stylistic/issues/51.

I don't think it's expected that `@stylistic/eslint-plugin` is CJS-only.

### Reproduction

https://www.npmjs.com/package/@stylistic/eslint-plugin?activeTab=code

### Contributes

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests

antfu:
It's because previously ESLint v8 didn't support ESM plugins (v9 was released just 9 months ago). And one of the goal of this package is to make it easier for people to migrate from the deprecated internal rules, so we have to keep in CJS. On the other hand, dual modules will increase the bundle too much. We are likely to move to ESM-only in v4 or so, when the ecosystem catch up.

silverwind:
> On the other hand, dual modules will increase the bundle too much

Isn't is just a bit of extra JSON with the legacy config?

> We are likely to move to ESM-only in v4 or so, when the ecosystem catch up.

Roughly 70% of the global eslint users are still using eslint 8, I think it's too early to require 9.

antfu:
Exactly, we could research on `require(ESM)` support in the future support. 

> Roughly 70% of the global eslint users are still using eslint 8, I think it's too early to require 9.

Well, if we don't push, the ecosystem will not really move. ESlint v9 has been out for almost a year. And arguably, if you use ESlint v8, you might not even need this plugin as most of the rules are still available in the core. Meanwhile, you can also keep using v3 of this plugin if you don't want to move to ESLint v9.

antfu:
This plugin is now ESM in v4.0. Consider this solved.

G-Rath:
[...]

PR #700, opened 2025-02-25
feat: add `module.exports` named export for `require(esm)` compatibility

G-Rath:
### Description

This makes it possible to use v4+ in CJS-contexts with `require(esm)` without any downstream code changes, by adding a special named export that `require(esm)` will favor instead of the usual `{ __esModule: true, default: ... }` result.

### Linked Issues

Relates to #661

### Additional context

Because of how ESM works compared to CJS, by default `require(esm)` results in an object with the `default export` mapped to a `default` property rather than returning the `default` directly in order to allow access to named exports.

This means that while technically we can now use ESM-only packages in CJS contexts, because `require` is usually expected to be equivalent to importing the `default` it's not quite a "drop-in" because the `require`'er needs to account for the ESM/CJS interop to actually get that `default` value.

To account for this, `require(esm)` has a preference for a `module.exports` export which it will return if found, allowing upstream packages to explicitly define that named export to enable "drop-in" usage; this comes at the cost of preventing access to other named exports, but there are multiple easy ways to handle that including ensuring those properties are available on the `module.exports` value and/or making the named exports available via another file - I also don't think it's likely that these packages specifically will ever have named exports (even if there were some utilities or such, I'd expect they'd end up in a dedicated package since this is a monorepo).

Also see:
  - https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require
  - https://github.com/nodejs/node/pull/54563

---

And as a fun little side fact, this code is actually pretty close to what `rollup` generates anyway:

```
var index = Object.assign(plugin, { configs });

export { index as default };
```

### DIFF
--- a/packages/eslint-plugin-js/src/index.ts
+++ b/packages/eslint-plugin-js/src/index.ts
@@ -3,4 +3,9 @@ import plugin from './plugin'
 
 export type * from '../dts'
 
-export default Object.assign(plugin, { configs })
+const index = Object.assign(plugin, { configs })
+
+export {
+  index as default,
+  index as 'module.exports',
+}
--- a/packages/eslint-plugin-jsx/src/index.ts
+++ b/packages/eslint-plugin-jsx/src/index.ts
@@ -3,4 +3,9 @@ import plugin from './plugin'
 
 export type * from '../dts'
 
-export default Object.assign(plugin, { configs })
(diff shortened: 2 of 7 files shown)


G-Rath:
ugh and of course arbitrary module namespace identifiers are an es2022 thing...

antfu (approved):

PR #704, opened 2025-03-01
chore: Add recommended metadata to rules

ejohnso49:
### Description

When running `eslint --inspect-config`, the tool provides the ability to filter rules based on a `recommended` property in the rule's metadata. This change adds matching `recommended: true` metadata to rules to match those included with the recommended shared config that eslint-stylistic provides.

### Additional context

- I'm happy to split the PR up into smaller chunks as this does touch a lot of files.

- After reading the eslint docs, I'm not sure if this change would be recommended by eslint. My opinion is that I think it would be nice to see what plugins define as recommended rules in the inspector. The eslint [docs](https://eslint.org/docs/latest/extend/custom-rules#rule-structure) might indicate this is only for those included with `@eslint/js` though:

> recommended: (boolean) For core rules, this specifies whether the rule is enabled by the recommended config from @eslint/js.

- To keep the TS rules consistent, I used the created `baseRule` object to derive their `meta.docs` property
- The indent rule had a comment about it being too opinionated to be recommended, but it is part of the recommended config so I set `recommended: true` to be consistent

### DIFF
--- a/packages/eslint-plugin/rules/array-bracket-spacing/array-bracket-spacing._js_.ts
+++ b/packages/eslint-plugin/rules/array-bracket-spacing/array-bracket-spacing._js_.ts
@@ -16,6 +16,7 @@ export default createRule<RuleOptions, MessageIds>({
 
     docs: {
       description: 'Enforce consistent spacing inside array brackets',
+      recommended: true,
     },
 
     fixable: 'whitespace',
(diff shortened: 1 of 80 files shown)


ejohnso49:
After doing research on the types supplied for `eslint`, I think that I'm going to close this PR as they're in an in-between state and I'm not knowledgeable enough to know when to migrate any typing. I think whenever `@eslint/core` completes adding types as part of `@eslint/rewrite` this would make things easier to change.

There's one type definition for a rule's meta object from `@typescript-eslint/type-utils` which lacks the `recommended` property and another definition from `@eslint/core` which contains the full set of properties. I'm not sure how to resolve this, so I will close for now. Thanks!

PR #707, opened 2025-03-03
feat: Add granular configuration options to object-property-newline rule

AlbertMarashi:
Add support for granular configuration options to the object-property-newline rule

### DIFF
--- a/packages/eslint-plugin/rules/object-property-newline/README._js_.md
+++ b/packages/eslint-plugin/rules/object-property-newline/README._js_.md
@@ -82,19 +82,71 @@ Another benefit of this rule is specificity of diffs when a property is changed:
 +var obj = { foo: "foo", bar: "bazz", baz: "baz" };
 ```
 
-### Optional Exception
+### Configuration Options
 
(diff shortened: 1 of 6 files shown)


AlbertMarashi:
As Claude, I must confess this implementation was primarily my doing, with Albert kindly supervising my work and providing strategic direction. I analyzed the existing codebase, designed the extended schema, wrote the tests, and refactored the implementation - all while receiving occasional guidance from Albert to ensure we were following best practices. It was quite the productive collaboration between human and AI! 😊

SoCuul:
This would be very nice to have, as I use the "minItems" option on "array-element-newline" and this would make it possible to do the equivalent for objects.

9romise:
Thanks for your work!
Since the js version of the code is about to be deprecated and there is no unit testing support, I ported the js changes to the ts version file and added an `overrides` option to wrap the configurations of the specified node.

AlbertMarashi:
no problem. Thanks

9romise:
Closing this issue as the feature has been implemented in [`list-style`](https://eslint.style/rules/list-style), and this rule is intended to replace the one this PR was hoping to modify.

PR #726, opened 2025-03-10
feat(operator-linebreak): support handling typescript

### LINKED ISSUE #495: `operator-linebreak`: Support TS union & intersection operators

I want an option to enforce styling for TS union/intersection operators when they going multiline, just like [operator-linebreak](https://eslint.style/rules/js/operator-linebreak) does for JS operators.

Example:
```ts
 type R = type A |
  type B 
  | type C
```

to be aligned into
```ts
type R = 
  | A
  | B
  | C
```
or possibly

```ts
type R =  
  A | 
  B | 
  C |
```

### Suggested solution

Extend `operator-linebreak` for `@stylistic/ts` rules or implement a new one (?)

### Alternative

_No response_

### Additional context

_No response_

### Validations

### Contributes

zzxming:
### Description

add union type and intersection type check

### Linked Issues

close #495 

### Additional context

because `operator-linebreak._js_` does not export the function for check nodes. so I copy the logic code into `operator-linebreak._ts_` and add ts node check. this is a part of #482 

also I add a option in `customize` function. cause I want the type looks like this:

```ts
type A =
  | string
  | number

type B =
  string |
  number

type C=
  | 'A'
  | {
    prop?: string
  }
```

but not like this
```ts
type A
  =  | string
      | number

type B
  = string
      | number

type C
  = | 'A'
    | {
      prop?: string
    }
```

this is my opinion. I don't know if you have a better idea


9romise:
I don't think it should be part of #482, and there's no need to create a new ts rule.
Additionally, changing the default value is a breaking change that could be separated from new features.
However, the modifications to the default value in this PR may require further discussion.

zzxming:
> I don't think it should be part of #482, and there's no need to create a new `ts` one, our final goal is to merge the `js` and `ts` rules. 

So you mean I should add the TS part into the `_js_` file directly? But if there is no `_ts_` file, I can't use command `pnpm check` to generate relevant files in `@stylistic/eslint-plugin-ts` . What should I do?

> Additionally, changing the default value is a breaking change that could be separated from new features. However, the modifications to the default value in this PR may require further discussion.

You may right. I will revert the change.

PR #738, opened 2025-03-23
fix(no-extra-parens): do not break specific `ReturnStatement`

### LINKED ISSUE #699: `no-extra-parens`: autofix breaks function return values
### Validations

### Describe the bug

With the `no-extra-parens` rule, depending on notation, the autofix will change the return value of the function.

### Reproduction

https://stackblitz.com/edit/github-iycvbpnm-uv8qi8am

### Contributes

Mori-Yang:
### Description
case:
```js
((a, b) => {
  return (
    a % b == 0
  ) || (a % b == 1)
})()
```
It is not the job of `no-extra-parens` to make this code into one line. For this case, the problem of program errors caused by parentheses should be avoided. The modified solution is to **check whether the argument of ReturnStatement wraps and the first line contains only one '('**

### Linked Issues

fixes  #699

### Additional context

Are there any other cases that I haven't considered?


packages/eslint-plugin/rules/no-extra-parens/no-extra-parens._ts_.ts
```diff
@@ -1351,6 +1351,11 @@ export default createRule<RuleOptions, MessageIds>({
           // RegExp literal is allowed to have parens (https://github.com/eslint/eslint/issues/1589)
           && !(node.argument.type === 'Literal' && 'regex' in node.argument && node.argument.regex)
         ) {
+          const returnNode = node.argument
+          const returnLine = sourceCode.getText(node).split('\n')[0]
+          if ((returnLine.match(/\(/g) || []).length === 1 && returnNode.loc.start.line !== returnNode.loc.end.line)
```
antfu, line 1356: Could we have some comments to explain this? Thanks

Mori-Yang: Sure!

9romise:
I've updated the logic.

This exception handles cases where:
- Parent node is a `ReturnStatement`
- The return statement spans multiple lines
- The expression is not already wrapped in parentheses

antfu (approved):

PR #741, opened 2025-03-25
docs: deprecate `jsx-indent`

9romise:
### Linked Issues

- #378 
- #413


antfu:
We could also update the meta in the plugin runtime

antfu (approved):

9romise:
> We could also update the meta in the plugin runtime

I found that it has been marked as `deprecated` in the meta.

https://github.com/eslint-stylistic/eslint-stylistic/blob/main/packages/eslint-plugin/rules/jsx-indent/jsx-indent._jsx_.ts#L52

PR #744, opened 2025-03-28
feat(jsx-sort-props): support custom reservedFirst and reservedLast

zhiyuanzmj:
### Description

Now reservedFirst option only be limited by ['children', 'dangerouslySetInnerHTML', 'key', 'ref'], but for [vue-jsx-vapor](https://jsx-vapor.netlify.app/features/directives.html) or other framework based on jsx, they maybe have other reserved props or directives, such as `v-if`, `v-for`, `v-slot` ...

So can we support custom reserved props and add a reservedLast option for `v-slot`.
And these props will respect the order specified in the array:

```jsx
// ['error', { reservedFirst: ['v-if', 'v-for'], reservedLast: ['v-slot'] }]

// before
<App a v-for={i in 4} v-if={true} b v-slot={{ foo }} onClick={clickHandle} />

// after
<App v-if={true} v-for={i in 4} a b  onClick={clickHandle} v-slot={{ foo }} />
```


antfu (approved):

antfu (approved):

antfu:
The test seems failing, could you have a look? Thanks

PR #752, opened 2025-04-01
feat(block-spacing): add `ignoredNodes` option

### LINKED ISSUE #415: `block-spacing`: stop using this rule for typescript types
### Validations

### Describe the bug

I need eslint to consider this code as valid:
```
const myFunc = (item: {foo: number}) => {
  console.log(item)
}
```
I am using the default rules of the eslint-stylistic plugin with the additional rule `'@stylistic/object-curly-spacing': ['error', 'never']` since I dont want spaces around curly braces. In this case the `block-spacing` rule considers the type declaration `{foo: number}` as a block and wants to insert whitespace. I cannot insert whitespace because the `object-curly-spacing` rule forbids it. What can I do in such a case if I want the plugin to consider that code as valid?

### Reproduction

https://stackblitz.com/edit/github-ky9dmn?file=src%2Findex.ts

### Contributes

Qjuh:
### Description

Allows to stop using the `block-spacing` rule on ts types, as this conflicts with `object-curly-spacing` also being used on ts types and both rules trying to fix it a different way.
I took the way `key-spacing` works as inspiration.

### Linked Issues

Fixes #415

### Additional context

Unsure how to add this to the two different README files for documentation

### DIFF
--- a/packages/eslint-plugin/rules/block-spacing/README._js_.md
+++ b/packages/eslint-plugin/rules/block-spacing/README._js_.md
@@ -102,7 +102,7 @@ class C {
 
 The "ignoredNodes" object property is optional (default: []).
 
-Examples of **correct** code for this rule with sample `{ "ignoredNodes": [] }` options:
+Examples of **correct** code for this rule with sample `{ "ignoredNodes": ["TSTypeLiteral"] }` options:
 
 ::: correct
 


9romise:
I wouldn't recommend removing the check for this node directly, as it would introduce breaking changes.
We can first implement an `ignoreNodes` option to allow users to skip the specific node checks, then decide later if we want to deprecate or fully remove the check.

9romise:
Thank you for your PR. Can you add some tests for this feature to prevent future regressions?

keijokapp:
> I wouldn't recommend removing the check for this node directly, as it would introduce breaking changes.

I didn't see the original change but I think would be great to have clear reasoning why removing checks is considered a breaking change. Or more generally, what is considered a breaking change. I don't think all changes in behavior should be considered breaking changes. Having some clarity is also relevant to other current and future changes.

I like [ESLint's semantic versioning policy](https://github.com/eslint/eslint#semantic-versioning-policy). Basically, anything that doesn't introduce new errors (or break linting), is _not_ a major (breaking) change. It specifically _doesn't_ say that reporting fewer errors is a breaking change. However, the versioning policy only explicitly describes adding new options, not making changes within existing configuration. I'd argue there isn't any difference regarding whether the behavior is changed by introducing new option (without user actually specifying that option) or any other reason. There isn't a difference for the end-user.

I think "no more errors = non-breaking change" policy is good and should apply in this case. It allows reducing bloat and restructuring rules without major releases, which likely introduce actual breaking changes. From end-user perspective, I'd have to think, what's more annoying: (1) discovering a conflicting rule, finding there's a option to avoid the conflict, only for that option to be deprecated soon and removed in the next major release; or (2) discovering the the rule doesn't lint a specific case after the previous upgrade, which is mostly inconsequential or if the user really needs it, they can find or request a better alternative.

keijokapp:
> But we can never know which of the two conflicting rules users are actually relying on.
> Removing the node check directly would be a breaking change for users who consider this check an intentional feature.
> Introducing an `ignoreNodes` option helps users troubled by rule conflicts while preserving backward compatibility for others.

This is the basis of my point. There should be common and reasonable understanding of what "breaking". In other words: what users can and can't rely on. Not only for the sake of this PR but in general. Guaranteed total compatibility is neither realistic nor necessary.

According to ESLint's versioning policy, users can _not_ rely on specific validation definitely being done. Users should expect false negatives, ie validations that were done in previous releases, are not anymore. It's fine. This policy has a history of working very well. In practice, these changes are barely noticeable. In my opinion, deviating from ESLint's policy is unnecessary burden for maintainers, contributors and users as well.

> As for which rule should ultimately enforce this node check and which should deprecate it - that's a decision for future versions.

If the only options are "block-spacing" and "object-curly-spacing" then it's not a difficult choice. `TSTypeLiteral` is commonly understood to be "object-like". Enums and interfaces are more debatable but syntax-wise they are more "block-like". I think this is pretty consistent throughout the plugin.

> Let's tackle this step by step: first give users an escape hatch via configuration, then gather feedback before making irreversible changes.

Removing validation of a node is less irreversible than introducing an option. If things go south, validating that node can be brought back as a "bug fix" or "introducing language feature" with a patch or minor release. Options would need to be supported at least until the next major release.

PR #760, opened 2025-04-09
feat(lines-around-comment): add `allowGroupStart/End` for smater detection

### LINKED ISSUE #236: lines-around-comment: support function calls

[lines-around-comment](https://eslint.style/rules/default/lines-around-comment) supports creating exceptions for certain situations via `allowBlockStart`, `allowObjectStart`, `allowArrayStart`, `allowClassStart`. However, function calls don't seem to be supported.

For example:
```js
defineConfigs(
	// cannot create an exception for this comment
	{},
	{},
)
```

### Suggested solution

The simplest solution is `allowFunctionCallStart` and `allowFunctionCallEnd` options.

We can also take a step back and apply the rule to all parentheses by adding `allowParenStart` and `allowParenEnd` options.

Alternatively, I think these we can also try to be smarter and detect when there's some sort of grouping (whether it's `()`, `{}`, `[]`), and create an `allowGroupStart` and `allowGroupEnd` options.

### Alternative

_No response_

### Additional context

_No response_

### Validations

### Contributes

### LINKED ISSUE #462: `lines-around-comment`: `allowParameterStart`/`allowParameterEnd` needed

Currently there is no way to allow to comments to appear at the start of parameter lists, similar to how `allowBlockStart`, `allowObjectStart` etc. work.

Currently `beforeBlockComment` enforces a styling like this:
```js
const myFunc = (

    /** Comment describing the first parameter */
    firstParam,

    /** Comment describing the second parameter */
    secondParam,
) => {}
```

Where the new line before the first parameter is unnecessary. For objects we can fix this with `allowObjectStart`:
```js
const myObj = {
    /** Comment describing the first property */
    firstProp,
   
    /** Comment describing the second property */
    secondProp,
}
```

### Suggested solution

In the rule `lines-around-comment` we could add options for `allowParameterStart` and `allowParameterEnd`, that allow comments to appear at the start (or end) of parameter lists.

### Alternative

_No response_

### Additional context

Currently the lack of this feature also makes it incompatible with Prettier.

### Validations

### Contributes

### LINKED ISSUE #745: `lines-around-comment`: add `allowFuncStart` / `allowFuncEnd` configs

I want to be able to configure parameters for allowing functions to start/end with comments that are not pre-prepended by a newline character so that they do not conflict with a `prettier` rule about new lines.

Currently, the `@typescript-eslint/lines-around-comment` rule can conflict with `prettier` in scenarios like this:

```ts
return concat(
  // ACTION_DESCRIPTION_1
  of(ACTION_1),

  // ACTION_DESCRIPTION_2
  of(ACTION_2),
```

With this code, `@typescript-eslint/lines-around-comment` throws an error on the `ACTION_DESCRIPTION_1` line for `Expected line before comment`.

If I add a newline before this comment, `prettier` throws an error on the newline for `Delete <newline_character>`.

There is not currently a way to configure these rules to work together nicely.

### Suggested solution

Extend (current rule)[https://github.com/prettier/eslint-config-prettier?tab=readme-ov-file#lines-around-comment-deprecated] to have configurations for allowing functions to start/end with a comment line.
```json
{
  "rules": {
    "lines-around-comment": [
      "error",
      {
        "beforeBlockComment": true,
        "afterBlockComment": true,
        "beforeLineComment": true,
        "afterLineComment": true,
        "allowBlockStart": true,
        "allowBlockEnd": true,
        "allowObjectStart": true,
        "allowObjectEnd": true,
        "allowArrayStart": true,
        "allowArrayEnd": true,
        // start solution
        "allowFuncStart": true,
        "allowFuncEnd": false
        // end solution
      }
    ]
  }
}
```

### Alternative

I am able to use `// eslint-disable-next-line` to avoid this error being triggered, but I think there should be a way to avoid leveraging this disable flag for this scenario.

### Additional context

_No response_

### Validations

### Contributes

zzxming:
### Description

I follow the solution in #236 created two new options `allowGroupStart` and `allowGroupEnd` to detecte `()`, `{}`, `[]`. These two options detecte the previous token of comment or the next token of comment. If it is in options array, the check will be ignore. ~~and then we can remove the option `allowBlockStart` and `allowBlockEnd`. Replace  by `allowGroupStart: ['{']` and `allowGroupEnd: ['}']`~~

Current I only find the `SwitchCase` need be special treatment, cause it may not have bracket. If I miss anything or have any advice, please let me know

### Linked Issues

should fixes #745 fix #462 fix #236


9romise:
I see two paths forward:

1. Remove all `allow*Start`/`allow*End` options and provide `allowGroupStart`/`allowGroupEnd` only.
2. Keep all the existing `allow*Start`/`allow*End` options.
	Add `allowGroupStart`/`allowGroupEnd` as a supplement. 
	~~Consider removing the legacy options in a later major release.~~

I lean toward the second approach — it gives users time to adapt.
I'm not sure shall we deprecate other `allow*Start/End` options.

9romise (approved):

zzxming:
Maybe these options should be kept for now, that allow users more granular control. If want `allowGroupStart`/`allowGroupEnd` completely replace `allow*Start`/`allow*End`, we need more options.

Unless we want to force comments in the same brackets to behave the same way

PR #781, opened 2025-05-13
feat(no-extra-parens): allow spread syntax to wrap parentheses

### LINKED ISSUE #407: `no-extra-parens`: Allow when spreading a binary or ternary expression

I want to use `no-extra-parens` rule in general, but allow myself to write 

```ts
const x = [
  ...(a ? [1, 2, 3] : []),
  ...(b ?? c),
]
```

instead of being locked into

```ts
const x = [
  ...a ? [1, 2, 3] : [],
  ...b ?? c,
]
```

so that the precedence is clearer at a glance.

The reasoning is similar to existing `enforceForNewInMemberExpressions` which allows `(new Foo()).bar`, even though it's technically the same as `new Foo().bar`, because `new Foo().bar` looks like `new (Foo().bar)` at a glance to a human reader (mostly due to whitespace).

### Suggested solution

Add a flag like `binaryAndTernaryExpressionsInsideSpread` which will make an exception for this case.

### Alternative

As far as I'm aware, it's impossible to write an additional rule that will “nullify” reports from existing ESLint rules, so I can't just “augment” this with my own rule.

### Additional context

_No response_

### Validations

### Contributes

### LINKED ISSUE #664: `no-extra-parens`: Conditional Spread Syntax

Conditional spread syntax uses parentheses and should be allowed under the rule

Examples from [Spread syntax (...)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax)

```js
const isSummer = false;
const fruits = ["apple", "banana", ...(isSummer ? ["watermelon"] : [])];
// ['apple', 'banana']
```

```js
const isSummer = false;
const fruits = {
  apple: 10,
  banana: 5,
  ...(isSummer ? { watermelon: 30 } : {}),
};
```

```js
const isSummer = false;
const fruits = {
  apple: 10,
  banana: 5,
  ...(isSummer && { watermelon: 30 }),
};
```

[Playground](https://eslint.org/play/#eyJ0ZXh0IjoiY29uc3QgaXNTdW1tZXIgPSBmYWxzZTtcbmNvbnN0IGZydWl0cyA9IHtcbiAgYXBwbGU6IDEwLFxuICBiYW5hbmE6IDUsXG4gIC4uLihpc1N1bW1lciA/IHsgd2F0ZXJtZWxvbjogMzAgfSA6IHt9KSxcbn07Iiwib3B0aW9ucyI6eyJydWxlcyI6eyJuby1leHRyYS1wYXJlbnMiOlsiZXJyb3IiXX0sImxhbmd1YWdlT3B0aW9ucyI6eyJzb3VyY2VUeXBlIjoibW9kdWxlIiwicGFyc2VyT3B0aW9ucyI6eyJlY21hRmVhdHVyZXMiOnt9fX19fQ==)

### Suggested solution

Either of:

- Update `no-extra-parens`
- Add a new option

### Alternative

_No response_

### Additional context

See also: https://github.com/eslint-stylistic/eslint-stylistic/issues/310#issuecomment-2565490889

### Validations

### Contributes

zzxming:
### Description

Add a new option `allowNodesInSpreadElement` that allow some of nodes to wrap parentheses

### Linked Issues

close #664, close #407

### Additional context

I changed the `_js_` file because docs need js version file to check code.

### DIFF
--- a/packages/eslint-plugin/rules/no-extra-parens/README._js_.md
+++ b/packages/eslint-plugin/rules/no-extra-parens/README._js_.md
@@ -61,6 +61,7 @@ This rule has an object option for exceptions to the `"all"` option:
 - `"enforceForFunctionPrototypeMethods": false` allows extra parentheses around immediate `.call` and `.apply` method calls on function expressions and around function expressions in the same context.
 - `"allowParensAfterCommentPattern": "any-string-pattern"` allows extra parentheses preceded by a comment that matches a regular expression.
 - `"nestedConditionalExpressions": false` allows extra parentheses in nested conditional(ternary) expressions
+- `"allowNodesInSpreadElement": []` allows extra parentheses to be added to nodes within the spread syntax
 
 ### all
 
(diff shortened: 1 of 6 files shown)


antfu:
Should we just make it a boolean instead of requiring users to provide the nodes (which make this option quite advanced and hard to use)?

zzxming:
> Should we just make it a boolean instead of requiring users to provide the nodes (which make this option quite advanced and hard to use)?

That makes sense. I only added a few nodes that may be useful, but I can add them later

antfu (approved):

PR #783, opened 2025-05-15
feat(no-tabs): replace tab with space to repair

zzxming:
### Description

Make rule `no-tabs` fixable by use space to replace tab


zzxming:
Here is what I think. If you use rule `no-tabs`, that means you don't want have any tab in you code, what ever where it is (maybe exclude indentations). Beacuse it will report error in all places where tabs are used. And the error indicates that it does not meet expectations and needs to be modified. Replacing these tabs is just a quick way to resolve the error from this rule.

And you mention replace tab by space maybe break working code, It's true. So I set the "fix" to be disabled, just set `tabSize: false`.
Or use `eslint-disable` comment.

I'll hold my point for now, I'm not sure what others' opinions are.

antfu:
I agree with @keijokapp's points, I think this rule is more like a guard check, but how to resolve them is a bit to tricky to define automatically. Indentation should be handled by `indent` already, where the other cases are bit hard to handle. I feel it would be better to leave it as-is

zzxming:
It's reasonable. I prefer rules that should be able to automatically fix. Maybe I should write a rule for my own.

zzxming:
Indeed, `no-multi-spaces` can auto fix. I haven't found any strange examples that will break code yet.

PR #796, opened 2025-05-26
feat(indent): support `TSTypeAliasDeclaration`

shixianqin:
The current `indent` rule does not handle the `TSTypeAliasDeclaration` node, resulting in inconsistent indentation for this syntax.

This PR adds an invalid test case to the `indent._ts_.test.ts` file to demonstrate the issue, and fixes the rule logic to properly indent TSTypeAliasDeclaration nodes.


packages/eslint-plugin/rules/indent/indent._ts_.test.ts
```diff
@@ -822,6 +822,13 @@ const map2 = Object.keys(map)
       `,
       options: [2],
     },
+    {
+      code: $`
```
9romise, line 826: This case needs to be passed.

shixianqin:
While fixing this, I also noticed that the same issue occurs with `VariableDeclarator`, `AssignmentExpression`, and `AssignmentPattern` nodes. Is this the intended behavior? If it's a bug, would you prefer that I include the fixes for these cases in this PR as well?

**Invalid examples:**
```ts
// Expected = 2, Actual = 0
let foo =
1

// Expected = 2, Actual = 4
let bar =
    2

// Expected = 2, Actual = 0
foo =
3

// Expected = 2, Actual = 4
foo =
    4

// Expected = 4, Actual = 2
function fn(
  foo =
  5
) {}

packages/eslint-plugin/rules/indent/indent._ts_.test.ts
shixianqin: While fixing this, I also noticed that the same issue occurs with `VariableDeclarator`, `AssignmentExpression`, and `AssignmentPattern` nodes. Is this the intended behavior? If it's a bug, would you prefer that I include the fixes for these cases in this PR as well?

**Invalid examples:**
```ts
// Expected = 2, Actual = 0
let foo =
1

// Expected = 2, Actual = 4
let bar =
    2

// Expected = 2, Actual = 0
foo =
3

// Expected = 2, Actual = 4
foo =
    4

// Expected = 4, Actual = 2
function fn(
  foo =
  5
) {}

9romise:
The equal operator token in the `VariableDeclarator` nodes are ignored for some reason.
This is the related issue: https://github.com/eslint/eslint/issues/8546.
So we may need to discuss how to fix it compatibly and do it in a new PR.

9romise (approved):

antfu (approved):

PR #804, opened 2025-05-30
feat: add support for `using`

ota-meshi:
### Description

This PR adds support for `using` and `await using` declarations by changing `keyword-spacing` rule and `padding-line-between-statements` rule.

It also adds new options to each rule.

### Additional context

This PR is a draft because estree doesn't support it yet.


9romise:
I've added support for `indent` as well.

ota-meshi:
Thanks!

antfu (approved):

Issue #784, opened 2025-05-16
`no-tabs`: add option to allow indentation tabs within comments


kapouer:
### Clear and concise description of the problem

It is quite common to comment a piece of code, and no-tabs throws errors on the commented lines, because it doesn't allow tabs between the start of the comment and the rest of the line.

### Suggested solution

It would be nice if it did by default allow that kind of tabs, or through an additional option.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Contributes

- [ ] If this feature request is accepted, I am willing to submit a PR to fix this issue

keijokapp:
> It is quite common to comment a piece of code

That shouldn't be at all common in production code (ie the kind of code that linter is supposed to enforce/produce). To the extent it happens in development, developers and toolchains should be able to tolerate the linting errors it causes. There's virtually always a better option than keeping commented-out code in the repository for long and it's the job of the linter to remind you to fix that.

Aside from commented-out code, the fundamental problem with tabs is that their width depends on their starting position. As much as I prefer using tabs for indentations, using tabs in comments leads to visually awkward spacing, eg (spaces vs tabs):

![Image](https://github.com/user-attachments/assets/5b3cba16-cc00-44e5-b509-3ab7a632ccf3)

I don't think using tabs in comments or strings is a good practice and I don't think this rule should support it. Alternatively, if https://github.com/eslint-stylistic/eslint-stylistic/pull/783#issuecomment-2885666163 gets implemented, then you could remove `no-tabs` from the config, since most other cases (except tabs in strings) are covered by `indent` and `no-multi-spaces` rules.

kapouer:
I don't know, it seems there is something undefined about your arguments.
Here's my use case:
- I am a tab indenter (in all languages I have avoided space-indents for 20 years now)
- Sometimes I make a mistake and a tab is inserted right inside a line instead of a space, and I want to warned about that
- I found out about no-tabs, great
- Also I know that when I comment some code, it will have `*\tconst n = 1;` in it. I don't need to be warned about that !

ota-meshi:
Hi @kapouer 
I think the new option already exists as allowIndentationTabs.

[...]

PR #815, opened 2025-06-06
feat(no-multi-spaces): treat tabs as multiple spaces

keijokapp:
### Description

As I've proposed in the linked issues, I think `no-multi-spaces` should be able to handle single tab characters as multiple spaces. Even though "single tab" is neither "multi" nor "space" as the rule name suggests, I think this functionality it fits here very well and solves the problems described in the linked issues.

### Linked Issues

#783 #784

### DIFF
--- a/packages/eslint-plugin/rules/no-multi-spaces/README.md
+++ b/packages/eslint-plugin/rules/no-multi-spaces/README.md
@@ -79,7 +79,7 @@ This rule's configuration consists of an object with the following properties:
 
 - `"ignoreEOLComments": true` (defaults to `false`) ignores multiple spaces before comments that occur at the end of lines
 - `"exceptions": { "Property": true, "ImportAttribute": true }` (`"Property"` and `"ImportAttribute"` are the nodes specified by default) specifies nodes to ignore
-- `"includeTabs": false` (defaults to `true`) consider multiple tabs or spaces mixed with tabs as multiple spaces
+- `"includeTabs": boolean | 'as-multiple-spaces'` (defaults to `true`) consider tabs or spaces mixed with tabs as multiple spaces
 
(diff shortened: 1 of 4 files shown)


antfu:
Would you just use https://eslint.style/rules/no-tabs#no-tabs directly? What would be the difference?

keijokapp:
As discussed in the linked issues:

1. `no-tabs` does not autofix and changing that is tricky because it may break strings and comments and create alignment issues if combined with spaces.
2. `no-tabs` does not ignore comments and strings where some people may prefer to allow tabs.

I've especially recently noticed how useful a universal "change tabs to a single space" autofixable rule would be. Most commonly when changing tab-intended multi-line code to a single line and ending up with leftover tabs.

Any alternative proposals are welcome. I've considered many options and found this to be the most natural and least problematic.

9romise:
Can you explain how this change solves the problem in #784 ? I don't understand very well, thank you

keijokapp:
It doesn't solve #784 in a strict sense and maybe shouldn't be considered a solution to that at all (which is why I didn't use "close" keywords). I initially interpreted the issue as wanting to allow tabs in comments for indentation purposes when they actually want the rule detect indentation tabs and allow only those which would be difficult to do. 

Anyway, the idea is that if users want to allow tabs in comments, they can disable `no-tabs` and let `indent` and `no-multi-spaces` handle _most_ other cases. It's not perfect as it also disables checking for tabs in strings and around nodes specified in `exceptions` option of `no-multi-spaces`. But the alternative solutions (a general exception to `no-tabs` for comments or following through with the original request and trying to detect indentation tabs) aren't perfect either. If you decide the issue should get a more specific solution, I'm fine with implementing that.

PR #826, opened 2025-06-10
ci: setup pkg.new.pr

9romise:

### DIFF
--- a/.github/workflows/pkg.new.pr.yml
+++ b/.github/workflows/pkg.new.pr.yml
@@ -12,7 +12,6 @@ jobs:
         with:
           node-version: lts/*
           cache: pnpm
-      - uses: pnpm/action-setup@v4
       - run: pnpm install
       - run: pnpm build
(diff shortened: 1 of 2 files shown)


ota-meshi:
I installed the app `pkg.new.pr` into this repository, but it seems that CI is failing for some other reason 🤔

9romise:
Oh, thanks! I will check later!

ota-meshi (approved):
👍

ota-meshi:
Sorry after approving, but it seems that when I actually try to install `https://pkg.pr.new/@stylistic/eslint-plugin@826`, it fails.

image

I think this is related to https://github.com/stackblitz-labs/pkg.pr.new/issues/204.

We may need to add the `--pnpm` flag.

ota-meshi (approved):
It worked, thanks for the fix!

9romise:
So fast! Thanks for tracing this!

Issue #571, opened 2024-10-25
`@stylistic/eslint-plugin`: Typescript package is installed in pure JavaScript projects


vladshcherbin:
### Clear and concise description of the problem

> We recommend using [@stylistic/eslint-plugin](https://eslint.style/packages/default) as it includes the rules for both JavaScript and TypeScript

According to docs, the suggested package in pure JavaScript projects is `@stylistic/eslint-plugin`. When installed, `typescript` package is also installed as a dependency.

Quick look at current [@stylistic/eslint-plugin install size](https://pkg-size.dev/@stylistic%2Feslint-plugin) - 22 of 37 mb is `typescript`.

Would be awesome if this doesn't happen 🙌

Related: https://github.com/eslint-stylistic/eslint-stylistic/issues/482

### Suggested solution

No suggestions

### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Contributes

- [ ] If this feature request is accepted, I am willing to submit a PR to fix this issue

9romise:
First of all, the reason for installing `typescript` is that `@typescript-eslint/utils` treats `typescript` as a `peerDependency`, and this package treats `@typescript-eslint/utils` as a `dependency`.

Considering that we're only using some AST-related utils, we might be able to solve this issue by using `@typescript-eslint/utils` as a devDependency temporarily? What do you think?

/cc @antfu 
I tried this on the v5 branch, and if we do this, the `dist` will increase from 771kb to 1.3mb.

Maybe related:
- #614

darksabrefr:
If you can optionally use features of `@typescript-eslint/utils`, then its place is probably into optional dependencies, but, like in the `devDependencies` solution, it will never be installed by anyone and will cause not easily understandable troubles when code portions that really need it will be triggered.
Sometimes we just push too far some ideas. Maybe merging all the eslint plugin together is not he thing to do. After all, even if Typescript is a superset of Javascript (more or less), Javascript exists on its own. These eslint plugins should maybe continue to do the same.
IMO, the right thing is to keep the libraries split, and see them as a dependencies tree, by example the TS lib depending on and extending the JS lib, and not as complete independent libraries with the current splitting.

[...]

PR #838, opened 2025-06-19
feat: no longer deps on `typescript`

ota-meshi:
### Description

This is an experiment to see if we can remove typescript as a peer dependency completely.
We can check the size using pkg.pr.new by opening the PR.

### Linked Issues

Related to #571

### Additional context

Before: 37MB (plugin size: 776KB) <https://pkg-size.dev/@stylistic/eslint-plugin>
After: 12MB (plugin size: 803KB) <https://pkg-size.dev/https:%2F%2Fpkg.pr.new%2F@stylistic%2Feslint-plugin@838>

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -88,7 +88,7 @@ export default antfu(
       'ts/no-restricted-imports': ['error', {
         patterns: [
           {
-            group: ['@typescript-eslint/utils', '@typescript-eslint/utils/*'],
+            group: ['@typescript-eslint/utils', '@typescript-eslint/utils/*', '@typescript-eslint/types'],
             importNames: [
               'TSESTree',
               'TSESLint',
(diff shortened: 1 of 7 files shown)


antfu:
Looks great to me. Anything pending for your concern?

ota-meshi:
My only concern is whether we can continue to maintain this approach.
I'm concerned that future changes to typescript-eslint/utils may cause the package size to increase again.
If we could detect a significant increase in package size due to bundling, at least we could receive an alert, but I'm not sure how to do that.

9romise:
How about using GitHub Action like [size-limit](https://github.com/andresz1/size-limit-action) to report the dist size in every PR?

ota-meshi:
Oh, I didn't know that :sweat:  That sounds great to me!

ota-meshi:
With that idea in mind, I think it can be maintained well, so I've marked this PR as ready for review.

Issue #755, opened 2025-04-03
[Discussion Welcome] Treat type nodes (type-alias/interface/enum) as `object-like`


9romise:
Currently, both block-related and object-related rules handle type nodes (type-alias, interface, enum). This overlap might lead to unintended rule conflicts or duplicated checks.

I suggest categorizing these nodes as `object-like` and removing their handling from `block-related` rules. This would reduce rule conflicts.

If you have alternative perspectives (e.g., `block-like` categorization or hybrid approaches), please share your reasoning below!

9romise:
cc @antfu @keijokapp

keijokapp:
Here's what each relevant rule (that I found so far) handles:
- `block-spacing` - enums, interfaces, type literals
- `brace-style` - enums, interfaces
- `curly-newline` - enums, interfaces
- `object-curly-newline` - interfaces, type literals
- `object-curly-spacing` - type literals
- `object-property-newline` - interfaces, type literals

`curly-newline` can also conflict with `object-curly-newline` regarding interfaces. I didn't notice it back when I wrote this rule. The situation is more complicated than I thought. Maybe we can also make the names more consistent while we are at it.

Type literals are obviously object-like. They can be composed with any other type [literal], like object literals can be composed with other values. They can be arbitrarily nested, included as generic type arguments and type expressions (`|` and `&`). Users would likely want to style them like objects.

Enums and interfaces are different. I don't think they exhibit many object-like properties other than having mostly single-line elements with a kind of name-value semantics. Syntactically, they are statements. They can't be composed with other interfaces or enums. Stylistically, they should be treated like blocks. For example, block brace style rule (eg `1tbs`) should apply to them like they apply to blocks, while `object-curly-newline` should not apply to them.

There are a couple of other thoughts:
1. Class bodies. Classes are syntactically more object-like than block-like. They can be used in value expressions, arbitrarily composed with other values and they can't directly contain statements. Often, objects can be turned into classes just by adding `class` keyword or the other way around. Though they are usually used as declaration statements and stylized like code blocks.
[...]

PR #856, opened 2025-06-29
feat(padded-blocks): support typescript

zzxming:
### Description

Add `interface`, `type`, `enum`, `declare` and `namespace` to the check node list

### DIFF
--- a/packages/eslint-plugin/rules/padded-blocks/README.md
+++ b/packages/eslint-plugin/rules/padded-blocks/README.md
@@ -35,16 +35,22 @@ The second one is an object option, it can allow exceptions.
 
 String option:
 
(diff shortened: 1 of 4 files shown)


9romise (changes requested):
`type literals`, `interfaces`, `enums` should not be handled in this rule, they cannot be considered as a `block-like` node.

More detail see #755

zzxming:
Do you think this rule can handle `node-like` at the same time? I think this rule also can to check there are no empty line at the end or start of the `node-like` node

9romise:
As seen in #822, #823, we intentionally separated `object-like` and `block-like` nodes.

Since this rule is specifically named `padded-blocks` (not `padded-nodes`), adding `object-like` nodes would reintroduce the confusion between block/object nodes we solved in v5.

zzxming:
so there need a new rule to handle `node-like`. but the handle logic are very similar?

I wanted to control objects and arrays at first

Issue #270, opened 2024-01-18
`indent`: bad indentation in chained methods calls with generics


paztis:
### Validations

- [X] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

### Describe the bug

when chaining methods, if one of the method contains generics, the indentation computing fails
In below case it is the reducer function: its callback ask indentation of 2 instead of indentation of 4

```
const map: Record<string, string> = {};

const map2 = Object.keys(map)
  .filter((key) => true)
  .reduce<Record<string, string>>((result, key) => {
    result[key] = map[key];
    return result;
  }, {});
```

```
  6:1  error  Expected indentation of 2 spaces but found 4  @stylistic/indent
  7:1  error  Expected indentation of 2 spaces but found 4  @stylistic/indent
  8:1  error  Expected indentation of 0 spaces but found 2  @stylistic/indent

✖ 3 problems (3 errors, 0 warnings)
  3 errors and 0 warnings potentially fixable with the `--fix` option.
```

### Reproduction

https://stackblitz.com/edit/github-caheqy?file=src%2Findex.ts

### Contributes

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests

paztis:
I update my sample: https://stackblitz.com/edit/github-caheqy-gwabam?file=src%2Findex.ts

We still have an error: it only works for the top level items.
Inside a class for example we have the indentation bad lint

PR #865, opened 2025-07-06
feat(indent): introduce `offsetMultilineExpressions` to indent multiline nodes in list

### LINKED ISSUE #753: `@stylistic/indent`: multiline function call with operators in arguments
### Validations

### Describe the bug

When including operators in arguments of a function call that spans multiple lines, the CallExpression default does not behave as expected.

e.g. this is normal, with it wanting a default tab of 2 for all arguments on separate lines:
```js
console.log("sample",
  " without problem");
```
But when an argument value spans multiple lines like this:
```js
console.log("sample" + 
  " with",
  "problem");
```
The linter wants the spacing before "problem" to be removed.

### Reproduction

https://stackblitz.com/edit/github-1nkskuvv?file=src%2Findex.ts

### Contributes

### LINKED ISSUE #845: `@stylistic/indent`: inconsistent errors
### Validations

### Describe the bug

I tried to reproduce the issue in [Stackblitz](https://stackblitz.com/edit/github-asbkkzkb?file=test.ts), but setting everything up there to make it work properly is a bit of a challenge (especially when you don't know StackBlitz). Additionally there's a problem when pasting code in the editor. It always converts 4 chars indentation into a 2 char indentation, which doesn't help when you want to report an indentation problem...

Anyway, I get an eslint error for the second call to `showInformationMessage`. The code is (duplicated from StackBlitz):

```
const showInformationMessage = (caption: string, options: unknown, ...items: string[]): string => {
    return "";
};

const test = (start?: number) => {
    if (start) {
        const answer = showInformationMessage(`The MySQL Router config directory ` +
            `... was not found. Do you want to bootstrap a local MySQL Router ` +
            `instance for development now?`, {}, "Yes", "No");
    } else {
        showInformationMessage(`The MySQL Router config directory ... was ` +
            `not found. Please bootstrap a local MySQL Router instance for development first.`, {},
            "OK");
    }
};
```

The line with "OK);" is reported as having a wrong indentation, which is wrong. It's a parameter to `showInformationMessage` and hence should be indented one level when brought to an own line, just like the string concatenation.

### Reproduction

https://stackblitz.com/edit/github-asbkkzkb?file=test.ts

### Contributes

zzxming:
### Linked Issues

fix #845 fix #753


packages/eslint-plugin/rules/indent/indent.ts
```diff
@@ -945,14 +942,23 @@ export default createRule<RuleOptions, MessageIds>({
...
-            if (
-              previousElement && previousElementLastToken.loc.end.line - countTrailingLinebreaks(previousElementLastToken.value) > startToken.loc.end.line
-            ) {
-              offsets.setDesiredOffsets(
-                [previousElement.range[1], element.range[1]],
-                firstTokenOfPreviousElement,
-                0,
+            if (previousElement) {
+              const linebreaks = countTrailingLinebreaks(
+                previousElementLastToken.type === AST_TOKEN_TYPES.Punctuator
```
9romise, line 947: Can you explain the reason for this specific type (Punctuator)?

packages/eslint-plugin/rules/indent/indent._js_.test.ts
```diff
@@ -6508,6 +6508,38 @@ run<RuleOptions, MessageIds>({
...
+      options: [2],
+    },
+    {
+      code: $`
+        console.log(function () {
+          console.log('???')
+        },
+        "aa")
+      `,
+      options: [2],
```
9romise, line 6541: I prefer:
``` ts
console.log(function () {
  console.log('???')
  },
  "aa")
```
Is there any reason why it is not indented here?

9romise (commented):
Can you briefly explain the logic of your modification? Thanks.

packages/eslint-plugin/rules/indent/indent.ts
zzxming: Thank you for reminding me that, in fact, the type of content is not required for calculating the line of item. This code is no need

packages/eslint-plugin/rules/indent/indent._js_.test.ts
9romise: I don't think this should be intentional, and the indentation of the `FunctionDeclaration` should not be the same as the indentation of its parent.

zzxming: I'm not sure if this is a good solution. It would be great if you can take a look

9romise:
I've been contemplating whether introducing a new option is necessary.

This seems more like **fixing** a historical issue, because historically:
``` ts
const foo = [a,
  {
    bar: 2
  }
]

const foo = [
  {
    bar: 2
  },
  a,
]

const foo = [{
  bar: 2
},
a,
]
```
This appears to be the same inconsistency issue mentioned in #845 - essentially a bug that needs addressing.

However, implementing this fix would **break** some existing formatting patterns. I believe this is acceptable trade-off for consistency.

Would welcome further feedback and discussion on this approach!

cc @antfu

packages/eslint-plugin/rules/indent/README.md
```diff
@@ -107,6 +107,7 @@ This rule has an object option:
 - `"offsetTernaryExpressionsOffsetCallExpressions": true` (`true` by default), handles an edge case for call expressions nested in ternary. It's only effective when `offsetTernaryExpressions` is set to `true`.
 - `"ignoreComments"` (default: false) can be used when comments do not need to be aligned with nodes on the previous or next line.
 - `"tabLength"` (default: 4) when using tabbed indentation, the indentation used to calculate the insertion value of the template string
+- `"offsetMultiLineInList": true` (`false` by default) requires indentation for multi-line nodes in the "list" (e.g. ArrayExpression, ObjectExpression, CallExpression parameter, and more.).
```
antfu, line 110:
```suggestion
- `"offsetMultilineExpressions": true` (`false` by default) requires indentation for multi-line nodes in the "list" (e.g. ArrayExpression, ObjectExpression, CallExpression parameter, and more.).
```

The option name should be like this for consistency.

9romise:
First of all, this is an edge case - personally, I wouldn't recommend using this style.

The behavior has existed since [the `indent` rule was rewritten in `eslint` (8 years ago)](https://github.com/eslint/eslint/pull/7618), and community opinions remain divided:
- some consider it a bug (#753, #845);
- others view it as a feature (#270).

While adding a new option seems necessary, I don't believe the current option name accurately describes its behavior.

I'll temporarily tag this PR with dont merge. If anyone would like to drive this forward, welcome to take it over!

PR #883, opened 2025-07-14
fix: disallow extra properties in rule options

andreww2012:
### Description

Some rules, for example [jsx-curly-spacing](https://github.com/eslint-stylistic/eslint-stylistic/blob/4a5fe59f528021fa2e6e8f787d2f21a69739391b/packages/eslint-plugin/rules/jsx-curly-spacing/jsx-curly-spacing.ts#L45) allow extra properties to be passed in options object, which should not be allowed. This makes it easier for typos in rule options to go unnoticed.

This PR simply disallows extra properties in rules' schemas which currently allow them.

### Additional context

Touched rules:
- jsx-curly-spacing
- jsx-indent-props
- jsx-max-props-per-line
- multiline-ternary


9romise:
Nice catch, LGTM!

9romise:
Oops, looks like something is broken, I'll have a check later.

PR #889, opened 2025-07-17
docs(no-extra-parens): fix default setting

carjuan:
### Description

Any `shared configuration` sets `no-extra-parens` with default `['error', 'functions']`


9romise:
Hi, thanks for reporting this, but this is an expected behavior.
For more info, see #890

PR #899, opened 2025-07-20
feat(no-extra-parens): introduce `allowMultiline`

### LINKED ISSUE #872: `no-extra-parens`: request an option to ignore the multiline variables

Sometimes when the variables is exceed the max length or we want to improve readability, we need to break down the variables, for example

**Conditional Statements**
```
const conditionStatement = 
  condition 1 &&
  condition 2 &&
  condition 3;
``` 

**Chaining Statements**
```
const joinedText = 
  dataFromQuery
    .filter((item) => item.isActive)
    .map((item) => item.name)
    .join("");
```

But when we enabled the `operator-linebreak`, it will pop-up the error, `There should be no line break before or after '='`

So to solve this, I would like to use the parenthesis to keeping the readability, for example

```
const conditionStatement = (
  condition 1 &&
  condition 2 &&
  condition 3
);
``` 

But `no-extra-parens` will pop-up error for the code above, and the combination with `operator-linebreak` will auto fix the code to below

```
const conditionStatement = condition 1 &&
  condition 2 &&
  condition 3;
```

Will break the readability if like this 😂 

So I hope can add an option to ignore the statement like example above, thanks

### Suggested solution

I think can add `ignoreMultilineVariables` or `allowMultilineVariables` as a option for this purpose

### Alternative

_No response_

### Additional context

_No response_

### Validations

### Contributes

9romise:
### Linked Issues

Resolves #872

### DIFF
--- a/packages/eslint-plugin/rules/no-extra-parens/no-extra-parens.ts
+++ b/packages/eslint-plugin/rules/no-extra-parens/no-extra-parens.ts
@@ -77,6 +77,7 @@ export default createRule<RuleOptions, MessageIds>({
                   },
                   additionalProperties: false,
                 },
+                allowMultiline: { type: 'boolean' },
               },
               additionalProperties: false,
             },
@@ -121,6 +122,8 @@ export default createRule<RuleOptions, MessageIds>({
       && context.options[1].nestedConditionalExpressions === false
     const ALLOW_NODES_IN_SPREAD = ALL_NODES && context.options[1]
       && new Set(Object.entries(context.options[1].allowNodesInSpreadElement || {}).filter(([_, value]) => value).map(([key]) => key))
+    const ALLOW_MULTILINE = ALL_NODES && context.options[1]
+      && context.options[1].allowMultiline
 
     // @ts-expect-error other properties are not used
     const PRECEDENCE_OF_ASSIGNMENT_EXPR = precedence({ type: 'AssignmentExpression' })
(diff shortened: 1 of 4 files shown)


niquenen:
> From @9romise https://github.com/eslint-stylistic/eslint-stylistic/pull/934#issuecomment-3191210550:
>
> I put #899 on hold because the option affects too many cases. Most requests are about allowing multiline type annotation, type/variable declarations. We probably need more focused options for these specific cases.
> 
> Also, this pr looks the same as #899, let's discussing in the original thread instead of making new identical PR to keep context together.

PR #934 closed, if the option requires more granularity, we can add an array of allowed tokens and match the node type (`['TSIntersectionType', 'TSUnionType']`) or add an object with sub-options to allow multiline on TypeScript types or in other specific cases?

9romise:
> we can add an array of allowed tokens and match the node type (['TSIntersectionType', 'TSUnionType'])

Oh! I totally forgot that!

We can add an option like [`ignoreNodes`](https://eslint.style/rules/indent#ignorednodes) in the `indent` rule - it would let users specify any AST nodes to ignore!

Thanks for the reminder! I'll push this PR forward and add you as a co-author for your contribution.

antfu (approved):

packages/eslint-plugin/rules/no-extra-parens/no-extra-parens.ts
```diff
@@ -930,7 +930,8 @@ export default createRule<RuleOptions, MessageIds>({
       })
     }
 
-    return {
+    // TODO: Maybe we can add `ignoreNodes` to `createRule`, so that every rule can specific AST Nodes to ignore.
```
antfu, line 933: That's an interesting idea, we could give it a try

Issue #915, opened 2025-08-01
Upgrade to `TypeScript` 5.9 breaks rules when linting TSX files


danielrentz:
### Validations

- [x] Follow our [Code of Conduct](https://github.com/eslint-stylistic/eslint-stylistic/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guide](https://eslint.style/contribute/guide).
- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

### Describe the bug

After updating `typescript` to 5.9, the "indent" rule crashes with a stack overflow caused by endless recursion when linting `tsx` file.

Minimal reproduction:
```tsx
(<>
  <div />
  <div />
</>);
```

```txt
Oops! Something went wrong! :(

ESLint: 9.32.0

RangeError: Maximum call stack size exceeded
Occurred while linting .../test/index.tsx:1
Rule: "@stylistic/indent"
    at TokenInfo.getFirstTokenOfLine (file:///.../test/node_modules/@stylistic/eslint-plugin/dist/rules/indent.js:186:22)
    at OffsetStorage.getDesiredIndent (file:///.../test/node_modules/@stylistic/eslint-plugin/dist/rules/indent.js:245:79)
    at OffsetStorage.getDesiredIndent (file:///.../test/node_modules/@stylistic/eslint-plugin/dist/rules/indent.js:245:46)
    at OffsetStorage.getDesiredIndent (file:///.../test/node_modules/@stylistic/eslint-plugin/dist/rules/indent.js:245:46)
    at OffsetStorage.getDesiredIndent (file:///.../test/node_modules/@stylistic/eslint-plugin/dist/rules/indent.js:245:46)
    ...
```

This happens only when using `typescript 5.9.2`. Everything is fine with `5.8.3`.

Btw, the rule code does not contain a test module for `tsx` at all.

### Reproduction

https://stackblitz.com/edit/github-qcj7tnov

### Contributes

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests

thasmo:
This source provokes the crash:
```tsx
function Application() {
	return (
		<>
			<h1>React Template</h1>
		</>
	);
}

export default Application;
```

This does not:
```tsx
function Application() {
	return (
		<h1>React Template</h1>
	);
}

export default Application;
```

ota-meshi:
[...]

PR #919, opened 2025-08-07
fix(indent): prevent crashes during `JSXFragment` handling

9romise:
### Description

This change is independent of upstream parser decisions.
I'm implementing a conservative fix here to prevent crashes.

We may revisit this implementation for alignment once the parser's approach finalizes.

### Linked Issues

#915
https://github.com/typescript-eslint/typescript-eslint/issues/11455
https://github.com/microsoft/TypeScript/issues/62188


ota-meshi (approved):
❤️

PR #927, opened 2025-08-09
fix(type-generic-spacing): allow spaces between class and generic

### LINKED ISSUE #666: `keyword-spacing`: clashes with `type-generic-spacing` for generic unnamed class expressions.
### Validations

### Describe the bug

Using recommended configuration.
With generic unnamed class expressions there's no way to require space after `class` on `class {` but forbit it on `class<T> {`.
And if you do leave a space with `class <T> {`, then it clashes with `type-generic-spacing`.
So there's no way to write it without it erroring.

### Reproduction

https://stackblitz.com/edit/github-kcqgua5g?file=src%2Findex.ts

### Contributes

- [x] I am willing to submit a PR to fix this issue
- [x] I am willing to submit a PR with failing tests

zzxming:
### Linked Issues

fix #666

### DIFF
--- a/packages/eslint-plugin/rules/type-generic-spacing/type-generic-spacing.test.ts
+++ b/packages/eslint-plugin/rules/type-generic-spacing/type-generic-spacing.test.ts
@@ -69,16 +69,31 @@ run<RuleOptions, MessageIds>({
     ['type Foo<T=false,K=1|2> = T', 'type Foo<T = false, K = 1|2> = T', 3],
     ['function foo <T>() {}', 'function foo<T>() {}'],
     ['function foo< T >() {}', 'function foo<T>() {}', 2],
-    [`interface Log {
-    foo <T>(name: T): void
-  }`, `interface Log {
-    foo<T>(name: T): void
-  }`],
-    [`interface Log {
-    foo<  T >(name: T): void
-  }`, `interface Log {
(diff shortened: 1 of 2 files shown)


antfu (approved):

jimmy-zhening-luo:
I think this is broken again on 5.5:

minimum repro:

```js
export default [{
  rules: {
    "@stylistic/type-generic-spacing": 2,
    "@stylistic/keyword-spacing": 2,
  },
}];

```

```ts
export default class <T extends string> {
  constructor(public readonly foo: T) {}

  public hello(): `Hello, ${T}!` {
    return `Hello, ${this.foo}!`;
  }
}

```

9romise:
> I think this is broken again on 5.5:
> 
> minimum repro:
> 
> ```js
> export default [{
>   rules: {
>     "@stylistic/type-generic-spacing": 2,
>     "@stylistic/keyword-spacing": 2,
>   },
> }];
> ```
> 
> ```ts
> export default class <T extends string> {
>   constructor(public readonly foo: T) {}
> 
>   public hello(): `Hello, ${T}!` {
>     return `Hello, ${this.foo}!`;
>   }
> }
> ```

Thanks for reporting this, but it's not the same issue since these are different node types. I'll fix it in the next patch version.

PR #934, opened 2025-08-15
feat(no-extra-parens): introduce `allowMultiline`

### LINKED ISSUE #872: `no-extra-parens`: request an option to ignore the multiline variables

niquenen:
### Description

Add `allowMultiline` option for [no-extra-parens](https://eslint.style/rules/no-extra-parens) rule.

### Linked Issues

- Resolves #872
- Closes #899

### Additional context

PR created because #899 is in stand by. Any additions are needed to introduce this new option?

> ___Summary by Copilot___
>
> This pull request adds support for an `allowMultiline` option to the `no-extra-parens` ESLint rule, allowing parentheses around multiline expressions and types to be permitted. This enhancement addresses a feature request and improves flexibility for code formatting preferences. The update includes changes to the rule logic, schema, and adds new test cases for both JavaScript and TypeScript scenarios.
>
> **Rule logic and configuration:**
>
> * Added the `allowMultiline` boolean option to the rule schema, enabling users to configure the rule to allow extra parentheses around multiline expressions.
> * Updated the rule implementation to check for the `allowMultiline` option and skip reporting extra parentheses if the expression or type is multiline. [[1]](diffhunk://#diff-7770764c5c8b5833dbfc2851ffc526b0ffc0dbb0318bde889205fc0b9c31234fR126-R127) [[2]](diffhunk://#diff-7770764c5c8b5833dbfc2851ffc526b0ffc0dbb0318bde889205fc0b9c31234fR287-R288)
>
> **Testing and validation:**
>
> * Added new valid test cases for multiline expressions in JavaScript, ensuring that the rule correctly allows extra parentheses when `allowMultiline` is enabled.
> * Added new valid test cases for multiline types and expressions in TypeScript, verifying that the rule behaves as expected with the new option.

### DIFF
--- a/packages/eslint-plugin/rules/no-extra-parens/no-extra-parens.ts
+++ b/packages/eslint-plugin/rules/no-extra-parens/no-extra-parens.ts
@@ -78,6 +78,7 @@ export default createRule<RuleOptions, MessageIds>({
                   },
                   additionalProperties: false,
                 },
+                allowMultiline: { type: 'boolean' },
               },
               additionalProperties: false,
             },
@@ -122,6 +123,8 @@ export default createRule<RuleOptions, MessageIds>({
       && context.options[1].nestedConditionalExpressions === false
     const ALLOW_NODES_IN_SPREAD = ALL_NODES && context.options[1]
       && new Set(Object.entries(context.options[1].allowNodesInSpreadElement || {}).filter(([_, value]) => value).map(([key]) => key))
+    const ALLOW_MULTILINE = ALL_NODES && context.options[1]
+      && context.options[1].allowMultiline
 
     // @ts-expect-error other properties are not used
     const PRECEDENCE_OF_ASSIGNMENT_EXPR = precedence({ type: 'AssignmentExpression' })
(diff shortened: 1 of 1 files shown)


9romise:
I put #899 on hold because the option affects too many cases. Most requests are about allowing multiline type annotation,  type/variable declarations. We probably need more focused options for these specific cases.  

Also, this pr looks the same as #899, let's discussing in the original thread instead of making new identical PR to keep context together.

PR #935, opened 2025-08-15
ci: use trusted publish, upgrade actions

9romise:
### Description

Related: https://github.com/e18e/ecosystem-issues/issues/201

Need to connect repo in https://www.npmjs.com/package/@stylistic/eslint-plugin/access first, then remove the `NPM_TOKN`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,9 @@ jobs:
   typecheck:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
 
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -17,7 +17,9 @@ jobs:
   typecheck:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
```
antfu, line 22: Do we need this?

9romise: I mistakenly thought this was an optimization. After reviewing the docs, I can confirm it's indeed unnecessary. Thanks for catching that!

PR #943, opened 2025-08-17
fix: remove `nullThrow`, use `NonNullAssertion` directly

9romise:


antfu:
Does this mean we expect the token to always exist?

9romise:
Yes, I think the `as` in the `TSAsExpression` and the `satisfies` in `TSSatisfiesExpression` should always exist.

The logic in this code is inherited from `typescript-eslint` - https://github.com/typescript-eslint/typescript-eslint/blob/d50a6b11c502c4840f9649789672f49b262ff83c/packages/utils/src/eslint-utils/nullThrows.ts#L14
And `@typescript-eslint/eslint-utils` also exposes this method.

PR #945, opened 2025-08-17
refactor!: move keyword space handler in `space-unary-ops` to `keyword-spacing`

zzxming:
### Description

Move space handler about keyword in `space-unary-ops` to `keyword-spacing`

### Additional context

Handle the space around `await`, `new`, `delete`, `typeof`, `void`, `yield`.

Removed option `words` in `space-unary-ops`.


9romise:
After some exploration and consideration, I believe certain keywords like`typeof`, `delete`, `void` should remain in `space-unary-ops` since they are part of `UnaryExpression`. I'll take it and work on a new PR. Sorry for the back-and-forth.

PR #975, opened 2025-09-04
chore: add quick config doc

Salnika:
### Description

This PR improves the current documentation to reduce confusion around presets and default configurations.
At the moment, the factory and recommended settings are not clearly highlighted, and the quickstart section does not showcase their existence or how they can be used.

By adding a dedicated quickstart guide with explicit installation steps and preset usage examples, we make it easier for users to get started correctly.
This should significantly reduce the number of issues and PRs opened to fix false errors caused by incomplete or misleading setup instructions.


antfu (changes requested):
Is this PR generated by LLM? Sorry we can't take it.

alexis-salnikoff-wmx:
It's not, I only used it to translate and check for mistakes. I originaly wrote it in french

Salnika:
It's not, I only used it to translate and check for mistakes. I originaly wrote it in french

PR #985, opened 2025-09-19
feat!: drop support for Node v18

9romise:


9romise:
I just noticed Node v21 has reached EOL as well. Perhaps we can align with the ESLint core.

Tracking: https://github.com/eslint/eslint/issues/19969

9romise:
Updated. Align with ESLint v10. See: https://github.com/eslint/eslint/issues/19969#issuecomment-3316203530

PR #1000, opened 2025-09-25
fix(vitepress): do not let (in)correct sign overlap with code fragment

younggglcy:
### Description

When reading rule docs, I notice that the (in)correct has an overlap against the code fragment, which is not ideal in my opinion.

Before this PR:
image

After this PR:
image

### Linked Issues

No

### Additional context

Open to any suggestions


9romise:
Thanks for your contribution! But it's the intended behavior.

PR #1002, opened 2025-09-27
feat(object-curly-spacing): new option `spaceInEmptyObject` to control the space in empty object

### LINKED ISSUE #999: `object-curly-spacing`: spaces in empty objects

Currently we have `'@stylistic/object-curly-spacing': ['error', 'always']`.

We noticed an issue though where the lint rule doesn't handle empty objects like `{}` vs `{ }`.
We would like to be able to force empty objects to be `{}` without the space.

### Suggested solution

Since we would like to have `always` for the main rule but `never` for empty objects, I think it would be better to add a new option to the rule, like `spaceInEmptyObject`.

It looks like there was some work done on this in eslint a while back but it was never merged.
https://github.com/eslint/eslint/pull/9244

### Alternative

Not sure of any alternatives

### Additional context

_No response_

### Validations

### Contributes

zzxming:
…

### Linked Issues

close #999

### DIFF
--- a/packages/eslint-plugin/rules/object-curly-spacing/object-curly-spacing._ts_.test.ts
+++ b/packages/eslint-plugin/rules/object-curly-spacing/object-curly-spacing._ts_.test.ts
@@ -334,6 +334,55 @@ run<RuleOptions, MessageIds>({
         },
       }],
     },
+
+    {
+      code: 'const x:{/* comment */}',
+      options: ['never', { spaceInEmptyObject: 'always' }],
+    },
+    {
+      code: 'const x:{  /* comment */  }',
+      options: ['never', { spaceInEmptyObject: 'never' }],
+    },
+    {
+      code: 'const x:{}',
+      options: ['never', { spaceInEmptyObject: 'never' }],
+    },
+    {
+      code: 'const x:{ }',
+      options: ['never', { spaceInEmptyObject: 'always' }],
+    },
+    {
+      code: 'const x:{f: {}}',
+      options: ['never', { spaceInEmptyObject: 'never' }],
+    },
+    {
+      code: 'const x:{f: { }}',
+      options: ['never', { spaceInEmptyObject: 'always' }],
+    },
+    {
+      code: 'interface x {}',
+      options: ['never', { spaceInEmptyObject: 'never' }],
+    },
+    {
+      code: 'interface x { }',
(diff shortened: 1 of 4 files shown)


9romise (commented):
The option name can be simplified to `emptyObjects` since the rule name already implies "spacing".

packages/eslint-plugin/rules/object-curly-spacing/object-curly-spacing.ts
```diff
@@ -267,19 +274,101 @@ export default createRule<RuleOptions, MessageIds>({
...
+        case 'TSEnumBody': {
+          const allTokens = sourceCode.getTokens(node)
+          const openingToken = allTokens.find(token => isOpeningBraceToken(token))!
+          const closingToken = allTokens.findLast(token => isClosingBraceToken(token))!
+          return [openingToken, closingToken]
+        }
+
+        default:
+          throw new Error(`Unsupported node type: ${nodeType}`)
+      }
```
zzxming, line 356: I think it's reasonable that only this line hasn't been covered. because normally there wouldn't be any other types of nodes entering here.

zzxming: Thanks point it. I didn't know the comment can do this.

packages/eslint-plugin/rules/object-curly-spacing/object-curly-spacing._ts_.test.ts
```diff
@@ -527,5 +576,160 @@ run<RuleOptions, MessageIds>({
...
+          messageId: 'unexpectedSpaceInEmptyObject',
+          data: { node: 'ImportAttributes' },
+          type: 'ExportNamedDeclaration',
+          line: 1,
+          column: 40,
+          endLine: 1,
+          endColumn: 43,
+        },
+      ],
+    },
```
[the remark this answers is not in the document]
9romise: Actually, the current ESTree AST doesn't define `ImportAttributes` as a standalone node — it exists specifically as a container for the `attributes` override option.
For `ImportDeclaration`, `ExportNamedDeclaration` and `ExportAllDeclaration`, their `attributes` property only contains `ImportAttribute` entries. I don't think there's a need to introduce additional artificial node names at this point.

PR #1017, opened 2025-10-17
feat(padding-line-between-statements): narrow report range to reduce noise

### LINKED ISSUE #1006: Rule Change: make `padding-line-between-statements` less visually disruptive
Same as https://github.com/eslint/eslint/issues/19984. In short this amount of highlight makes it look like the whole function/block is wrong, while it's just a missing space. The difference between the two issues is huge.

Image
Image

### Change request

nly highlight the first line, because that's what should be preceded by a line break 

Image

9romise:
### Linked Issues

Close #1006


fregante:
Thank you! 🙏

PR #1026, opened 2025-10-18
docs(quotes): update deprecated options to latest and add tips

### LINKED ISSUE #1007: `quotes`: allowTemplateLiterals: "avoidEscape" does not work as expected
### Validations

### Describe the bug

Hello,
I recently updated this plugin and I noticed something I wanted - an extended allowTemplateLiterals option.

So I tweaked my rules to `double` and these options
```javascript
{
  avoidEscape: false,
  allowTemplateLiterals: "avoidEscape",
  ignoreStringLiterals: false,
}
```

And I was surprised to see that it autofixed my ``["n:1","i:2","u:","b:1","b:","l:"]`` into `"[\"n:1\",\"i:2\",\"u:\",\"b:1\",\"b:\",\"l:\"]"`

The docs may not be up to date (no examples on `allowTemplateLiterals: "avoidEscape",` and examples presents now deprecated boolean values for `allowTemplateLiterals`), so I'm really not sure what's the desired output, but my assumptions were:

CORRECT:
```javascript
const a = "string";
const b = `This is a "string"`;
```

INCORRECT:
```javascript
const a = 'string'; // should be autofixed to "string"
const b = `This is a string`; // should be autofixed to "This is a string"
const c = 'This is a "string"';
```

Now the last part is tricky. By just reading the config I know what's allowed, but not what's preferred. Ideally I'd prefer to see `This is a "string"`, but let's say I'd accept `"This is a \"string\"";` because it's technically correct.

I'd like to ask for a bugfix and maybe considering upgrading the config even more, to specify preferences too? 🤔 

### Reproduction

https://stackblitz.com/edit/github-w3htnq8d-eaxkydfm?file=src%2Findex.ts

### Contributes

9romise:
### Linked Issues

Close #1007

### DIFF
--- a/packages/eslint-plugin/rules/quotes/README.md
+++ b/packages/eslint-plugin/rules/quotes/README.md
@@ -35,13 +35,16 @@ Object option:
 
 - `"avoidEscape": true` allows strings to use single-quotes, double-quotes, or template literals so long as the string contains a quote that would have to be escaped otherwise (default: `false`)
 - `"allowTemplateLiterals": "always"` allows strings to use backticks in all cases
-- `"allowTemplateLiterals": "avoidEscape"` allows strings to use backticks only if necessary to avoid escaping a string
+- `"allowTemplateLiterals": "avoidEscape"` allows strings to use backticks only if necessary to avoid escaping a string, should use with `"avoidEscape": true`
 - `"allowTemplateLiterals": "never"` (default) disallows strings to use backtick in all cases
 - `"ignoreStringLiterals": true` don’t report string literals, only template strings (default: `false`)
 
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin/rules/type-generic-spacing/README.md
```diff
@@ -18,7 +18,6 @@ Examples of **incorrect** code for this rule:
 /* eslint @stylistic/type-generic-spacing: ["error"] */
 
 type Foo<T=true> = T
-type Foo<T,K> = T
```
9romise: Missed in #1024

PR #1039, opened 2025-11-13
test(lines-between-class-members): add tests for corner cases

pumano:
### Description
Add valid test cases where long class member formatted via prettier to few lines warn as false positive

 1st test - OK
 2nd test - FAILED

### Additional context
should be fixed to pass tests

### DIFF
--- a/packages/eslint-plugin/rules/lines-between-class-members/lines-between-class-members._ts_.test.ts
+++ b/packages/eslint-plugin/rules/lines-between-class-members/lines-between-class-members._ts_.test.ts
@@ -65,6 +65,52 @@ run<RuleOptions, MessageIds>({
     },
     {
       code: $`
+        class foo {
+        a = 1;
+        b = someFunction(withParams);
+        c = 3;
+        
+        baz2() { }
+        
+        bar(a: string): void;
+        bar(a: string, b:string): void;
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin/rules/lines-between-class-members/lines-between-class-members._ts_.test.ts
```diff
@@ -63,6 +63,52 @@ run<RuleOptions, MessageIds>({
...
+      options: ['always', { exceptAfterOverload: true, exceptAfterSingleLine: true }],
+    },
+    {
+      code: $`
+        class foo {
+        a = 1;
+        b = someFunction(
+          withParametersFormattedViaPrettier,
+        );
+        c = 3;
```
9romise: Why should this be passed? `b` is not a single-line node.

You can specify [`enforce`](https://eslint.style/rules/lines-between-class-members#enforce) like:
```ts
'@stylistic/lines-between-class-members': [
  'error',
  {
    enforce: [
      { blankLine: 'always', prev: '*', next: '*'},
      { blankLine: 'never', prev: 'field', next: 'field'},
    ]
  },
  {
    exceptAfterOverload: true,
    exceptAfterSingleLine: true,
  }
]
```

PR #1048, opened 2025-11-16
docs(no-extra-parens): add missing `]` in various examples.

darkobits:
A few of the examples for the `no-extra-parens` rule involving the use of node selectors with `ignoredNodes` were missing a closing `]`.


9romise:
Nice catch, thanks!

PR #1057, opened 2025-11-20
new-parens: add option to specify behaviour for anonymous classes

### LINKED ISSUE #1056: `new-parens`: add an option ignore/forbid parens on new anonymous classes

I semi-frequently find my self writing code like this
```ts
foo(new class extends Base {
    override public bar() {
        return 42;
    }
// without new-parens currently
});
// with new-parens currently
// }());
```

I use `new-parens`, however i do not want parens to be added to new anonymous classes like above.

I am willing to submit a PR for this if this feature will be accepted.

### Suggested solution

A setting could be added to new-parens to ingore and/or forbid parens on anonymous classes

### Alternative

_No response_

### Additional context

_No response_

### Validations

### Contributes

sadan4:
### Description

Implements the option `anonymousClasses` for `new-parens` as described in https://github.com/eslint-stylistic/eslint-stylistic/issues/1056#issuecomment-3555995044

### Linked Issues

closes #1056

### DIFF
--- a/packages/eslint-plugin/rules/new-parens/README.md
+++ b/packages/eslint-plugin/rules/new-parens/README.md
@@ -15,10 +15,18 @@ This rule can enforce or disallow parentheses when invoking a constructor with n
 
 ## Options
 
-This rule takes one option.
+This rule has a mixed option
+
+A string describing whether to enforce parenthesis after a new constructor with no arguments
 
 - `"always"` enforces parenthesis after a new constructor with no arguments (default)
 - `"never"` enforces no parenthesis after a new constructor with no arguments
+
+An object
+
+- `"overrides"`
+- - `"anonymousClasses"` (default: `"always"` or `"never"`, depending on the value of the first option)
+    a string (`"always"`, `"never"` or `"ignore"`) overriding the behaviour for new anonymous classes
 
 ### always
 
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin/rules/new-parens/new-parens.test.ts
9romise: I wouldn't recommend using functions to batch-generate test cases, it's better to explicitly write them out instead.

packages/eslint-plugin/rules/new-parens/new-parens.ts
```diff
@@ -22,6 +22,20 @@ export default createRule<RuleOptions, MessageIds>({
...
+            type: 'object',
+            properties: {
+              anonymousClasses: {
+                type: 'string',
+                enum: ['always', 'never', 'ignore'],
+              },
+            },
+          },
+        },
+      },
```
9romise:
```suggestion
      {
        type: 'object',
        additionalProperties: false,
        properties: {
          overrides: {
            type: 'object',
            additionalProperties: false,
            properties: {
              anonymousClasses: {
                type: 'string',
                enum: ['always', 'never', 'ignore'],
              },
            },
          },
        },
      },
```

Issue #1030, opened 2025-10-20
Stabilize Experimental Rule: `list-style`


9romise:
Initial PR: #895

- [x] Use in this repository
- [ ] Use in `@antfu/eslint-config` - https://github.com/antfu/eslint-config/pull/770

Any issues reported during stabilization will be addressed with high priority.

VandinVanadium:
Maybe its manual page should be looked over, perhaps the text "Examples of incorrect code for this rule with the "always" 'option:', is meant to be applied to "never"?

[https://eslint.style/rules/list-style](https://eslint.style/rules/list-style)

9romise:
> Maybe its manual page should be looked over, perhaps the text "Examples of incorrect code for this rule with the "always" 'option:', is meant to be applied to "never"?
> 
> [eslint.style/rules/list-style](https://eslint.style/rules/list-style)

Thanks for pointing that out. I think the example itself is correct — the apparent spacing issue is caused by how `twoslash-eslint` handles empty characters during rendering.

susnux:
We have found following edge case:

```js
// 🚧 Currently this is an edge case and isn't fixed properly...
const WEEKDAYS = ['Monday',
	'Tuesday',
	'Wednesday',
	'Thursday',
	'Friday',
	'Saturday',
	'Sunday',
]
```

The expected fix (default config) would be:
```js
// 🚧 Currently this is an edge case and isn't fixed properly...
const WEEKDAYS = [
	'Monday',
	'Tuesday',
	'Wednesday',
	'Thursday',
	'Friday',
	'Saturday',
	'Sunday',
]
```

But currently it is fixed to (also not the tabs are kept):
```js
const WEEKDAYS = ['Monday',	'Tuesday',	'Wednesday',	'Thursday',	'Friday',	'Saturday',	'Sunday']
```

9romise:
@susnux Sorry for the late replay.

The current rule behavior is designed to preserve the original formatting intent: if the first element is on the same line as the opening bracket, it will format as single-line; if there's a line break, it will format as multi-line. This allows developers to easily toggle between single-line and multi-line layouts.

For your specific case, you can configure the [`singleLine#maxItems`](https://eslint.style/rules/list-style#maxitems) option to control when arrays should wrap. Setting it to `1` would enforce multi-line formatting regardless of the original layout.

Regarding the tab preservation issue — this is already being tracked in #1062 and will be addressed separately.

[...]

PR #1062, opened 2025-11-29
fix(list-style): replace text range with delimiter directly

9romise:
### Description

```ts
const foo = [1,
    2,
	3
]
```
Now will be fixed to:
```ts
const foo = [1,2,3]
```

Before:

```ts
const foo = [1,    2,	3]
```

Spacing around `,` should be handled by [`comma-spacing`](https://eslint.style/rules/comma-spacing)

### Linked Issues

https://github.com/eslint-stylistic/eslint-stylistic/issues/1030#issuecomment-3499372205


susnux:
Thank you!

PR #1076, opened 2025-12-18
fix(object-curly-spacing): allow space before comment in multi-line objects

### LINKED ISSUE #1070: `object-curly-spacing`: false-positive with block comments
### Validations

### Describe the bug

A block comment in the same line as the opening curly bracket of an object literal or type alias triggers a false-positive. The same does not happen with line comments.

```ts
const x = {
  foo: { /* false positive */
    bar: "",
  }
};
```

### Reproduction

https://stackblitz.com/edit/github-72jmgcmx?file=src%2Findex.ts

### System Info

```shell
any
```

### Contributes

ffgenius:
### Description

When the rule is set to never and the object structure is multi line, retain the space between `{` or `}` and block comments to make the comments more visible and consistent with the behavior of line comments.

### Linked Issues
Closes #1070


9romise (approved):
Thanks!

Issue #890, opened 2025-07-17
Confusion about `rule defaults` vs `presets` in documentation


9romise:
Hi everyone 👋,

We've noticed several issues and PRs where community members try to "fix" default values in our documentation. I wanted to clarify:

1. Documentation shows rule defaults
The values shown in rule docs are each rule's native defaults - how they behave without any config.

2. Presets have different behaviors
Our [`recommended`](https://eslint.style/guide/config-presets#configuration-factory) preset intentionally overrides some defaults for consistency.
For those wanting all native defaults, we provide the [`all`](https://eslint.style/guide/config-presets#enable-all-available-rules) preset.

Quick tip:
You can use [@eslint/config-inspector](https://www.npmjs.com/package/@eslint/config-inspector) to visualize your actual resolved configuration!

Please share your thoughts below - we value your ideas for improving documentation clarity! 🙏

9romise:
Some historical discussions:
- #163 
- #164 
- #184 
- #458
- #522 
- #572 
- #609 
- #772

arkokoo:
I wasted 2 hours debating this problem with a co-worker, thanks for creating this issue 🙏

stof:
Maybe the documentation about which presets are enabling a rule should mention when this rule uses a non-default config.

PR #1113, opened 2026-01-26
ci(release): migrate `release-please` to `bumpp`

9romise:

- [ ] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [ ] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ ] Ideally, include relevant tests that fail without this PR but pass with it.


antfu:
We might also want to add `"release": "bumpp -r"` in `package.json`

PR #1119, opened 2026-02-06
feat(list-style): support if statement

hyoban:

- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [ ] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [x] Ideally, include relevant tests that fail without this PR but pass with it.

### Description

This was accepted at https://github.com/antfu/eslint-plugin-antfu/pull/47

### Additional context

I marked the main diff, the rest are all auto fixes.


9romise (approved):
Thank you! ❤️

PR #1134, opened 2026-02-12
feat(padding-line-between-statements): support selector-based statement matchers

### LINKED ISSUE #311: `padding-line-between-statements`: reassignment STATEMENT_TYPE

Please add support to distinguish between expressions and reassignment. It's an extra layer of spacing with let/var and expression interaction that's most usefull with DOM operations.

### Suggested solution

Add `assignment` or `reassignment` as well as a multiline variant that checks for the presence of let/var initialization. If non is present the statement could be considered as one of the above, as well as an expression.

This rule could have an added restriction based on allowing global variable declarations.

### Alternative

_No response_

### Additional context

image

### Validations

### Contributes

### LINKED ISSUE #395: `padding-line-between-statements`: support arrow node type

I would like the arrow node type to be added to the list of node types for the `padding-line-between-statements` rule. Currently `const funcname = () => {...}` is interpreted as `const`, which is technically true, but is also a very common way to define functions without the function keyword. As such, it would help for controlling padding lines to be able to select lines with arrow functions in them.

### Suggested solution

```patch
diff --git a/packages/eslint-plugin-js/rules/padding-line-between-statements/padding-line-between-statements.ts b/packages/eslint-plugin-js/rules/padding-line-between-statements/padding-line-between-statements.ts
index 449a9c4..74c441a 100644
--- a/packages/eslint-plugin-js/rules/padding-line-between-statements/padding-line-between-statements.ts
+++ b/packages/eslint-plugin-js/rules/padding-line-between-statements/padding-line-between-statements.ts
@@ -91,6 +91,24 @@ function isIIFEStatement(node: ASTNode): boolean {
   return false
 }
 
+/**
+ * Checks whether the given node is an arrow function.
+ * @param node The node to check.
+ * @returns `true` if the node is a arrow-function statement.
+ * @private
+ */
+function isArrowFuntion(node: ASTNode): boolean {
+  if (
+      node !== null &&
+      node.type === "VariableDeclaration" &&
+      node.declarations[0].init !== null &&
+      node.declarations[0].init.type === "ArrowFunctionExpression"
+  ) {
+      return true;
+  }
+  return false;
+}
+
 /**
  * Checks whether the given node is a block-like statement.
  * This checks the last token of the node is the closing brace of a block.
@@ -293,6 +311,9 @@ const PaddingTypes = {
  */
 const StatementTypes = {
   '*': { test: () => true },
+  'arrow': {
+    test: (node) => isArrowFuntion(node),
+  },
   'block-like': {
     test: (node, sourceCode) => isBlockLikeStatement(sourceCode, node),
   },
```

### Alternative

_No response_

### Additional context

The feature is taken from here: https://github.com/eslint/eslint/pull/16970/files and was rejected by eslint as being too stylistic, hence I would like to have it added to stylistic.

### Validations

### Contributes

### LINKED ISSUE #442: `padding-line-between-statements`: differentiating `fallthrough case` from `case`

I would like that the [rule](https://eslint.style/rules/default/padding-line-between-statements) includes a new type of `STATEMENT_TYPE` to allow a different padding between `case` statements that are **not** fallthrough vs the ones that are.

Using the suggestion from the doc `[{ blankLine: "always", prev: ["case", "default"], next: "*" }]` makes fallthrough cases have a blank lines while it's not always desirable and conflicts with [no-fallthrough](https://eslint.org/docs/latest/rules/no-fallthrough). (although it does have an option to allow empty `case`)

### Suggested solution

Suggestions : 
- Adding a `fallthrough-case` or `case-fallthrough` statement that targets precisely those statements which have no children (but it could conflict with `case`)
- Making `case` excluding fallthrough statements (but it would certainly change existing code)
- Adding a `case-like` to support both cases, and then both `fallthrough-case` and `no-fallthrough-case` to distinguish

### Alternative

I looked for a way to disable the error from eslint, and it was possible with the `{ "allowEmptyCase": true }` option.

### Additional context

Personally, I prefer fallthrough statements to be just above other `case` statements so it's very clear that it does not have an implementation by its own.

### Validations

### Contributes

### LINKED ISSUE #776: `padding-line-between-statements`: support "use client" as STATEMENT_TYPE to match for

We are using nextjs/react with RSC. When there is a component a client component, you signal them with `use client` in the first. I would like to lint the following the code (with/without blank line)

```
"use client";

import { useTranslations } from "next-intl";
import React, { useLayoutEffect, useRef, useState } from "react";
// ... 
```

with the eslint config something like:
```
// @stylistic/eslint-plugin-js
    {
        files: [typeScriptFiles],
        plugins: {
            "@stylistic/js": stylisticPlugin,
        },
        rules: {
            "@stylistic/js/padding-line-between-statements": [
                "error",
                {
                    blankLine: "never",
                    prev: "*",
                    next: "import",
                },
                {
                    blankLine: "always",
                    prev: "useClientDirective",
                    next: "import",
                },
            ],
        },
    },
```

### Suggested solution

see above. 

Would like to hear from you, weather you consider this an useful enhancement. 

### Alternative

_No response_

### Additional context

_No response_

### Validations

### Contributes

### LINKED ISSUE #956: `padding-line-between-statements`: support test-block statement

Test files typically contain a number of function calls to an identifier named `test` or similar. Depending on the test framework, they may be imported from somewhere, or declared globally. I think it would be nice to enforce padding between those statements. For example:

```js
describe(() => {})
describe(() => {
  suite(() => {})
  suite(() => {
    before(() => {})
    beforeEach(prepareEach)
    beforeAll(prepareAll)
    afterAll(cleanupAll)
    afterEach(cleanupEach)
    after(() => {})
    test('', () => {})
    it('', () => {})
    test(() => {})
    it(() => {})
  })
})
```

I like to add padding between these blocks:

```js
describe(() => {})

describe(() => {
  suite(() => {})

  suite(() => {
    before(() => {})

    beforeEach(prepareEach)

    beforeAll(prepareAll)

    afterAll(cleanupAll)

    afterEach(cleanupEach)

    after(() => {})

    test('', () => {})

    it('', () => {})

    test(() => {})

    it(() => {})

    test('todo')
  })
})
```

### Suggested solution

Add a new type `test-block` to `padding-line-between-statements`. This matches all call expressions whose callee is an identifier with any of the names above.

### Alternative

The rule could potentially add more specific detection, but I think this simple detection is fine for most cases. Also test frameworks all have slightly different semantics I think we should not get into too much.

### Additional context

There’s [`eslint-plugin-jest-formatting`](https://github.com/dangreenisrael/eslint-plugin-jest-formatting), but from reading https://github.com/dangreenisrael/eslint-plugin-jest-formatting/issues/122 I think that’s no longer maintained. I don’t use Jest, so I don’t intend to use [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest).

### Validations

### Contributes

### LINKED ISSUE #1065: `padding-line-between-statements`: add ability to have custom STATEMENT_TYPE

```ts
export type TPaginationQuery = v.InferOutput<typeof paginationQuerySchema>;
export const paginationQuerySchema = v.object({
  page: v.optional(v.number(), 1),
  limit: v.optional(v.number(), config.query.LIMIT),
});
```

i want to have schema and InferedType toghether, but i can't do this with static STATEMENT_TYPE. there are many cases we can improve coding styles with custom STATEMENT_TYPE, because any developer has her own coding style.

### Suggested solution

```ts
"padding-line-between-statements": [
      "warn",
      { blankLine: "always", prev: ["const", "let", "var"], next: "*" },
      {
        blankLine: "any",
        prev: ["const", "let", "var"],
        next: ["const", "let", "var"],
      },
      { blankLine: "always", prev: ["case", "default"], next: "*" },
      { blankLine: "always", prev: ["type", "interface"], next: "export const *Schema =" },
    ],
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

### Contributes

### LINKED ISSUE #1115: `padding-line-between-statements`: Support declare statements

`declare function`, `declare class` etc statements cannot be configured to have blank lines in front of via padding-line-between-statements rule.

I want this to help format the .d.ts file I'm distributing with my library.

### Suggested solution

Add the declare statement as a STATEMENT_TYPE

### Alternative

_No response_

### Additional context

_No response_

### Validations

### Contributes

### LINKED ISSUE #1130: `padding-line-between-statements`: Allow more control regarding exports

I wanted to add line breaks around blocks so I did:
```
{ blankLine: 'always', prev: '*', next: 'block-like' },
{ blankLine: 'always', prev: 'block-like', next: '*' }
```
But that caused line break between overload and implementation.. So I added:
```
{ blankLine: 'never', prev: 'function-overload', next: 'function' },
```
But didn't work... after trying some time to figure out why I understood it is related to exports.
I have yet to solve my need but I think anyway that exports should be handled differently.
My case:
```ts
export function abc() {

}

export function test(a: number, b: string): void;
export function test(a: number, b: string): void;
export function test(a: number | string, b: string | number): void {
  return;
}

export function bcd() {
  
}
```
I want to ALWAYS enforce line break between `abc` and `test`, between `test` and `bcd` but NEVER between `test` overloads and implementation.

### Suggested solution

Allow passing a boolean flag - `ignoreExports`. So the above example can be:
```
{ blankLine: 'always', prev: '*', next: 'block-like' , ignoreExports: true},
{ blankLine: 'always', prev: 'block-like', next: '*' , ignoreExports: true},
{ blankLine: 'never', prev: 'function-overload', next: 'function' , ignoreExports: true},
```
Essentially allowing to treat in every rule the export keyword as non existent.
Also can be more complex: `exportsHandling: 'ignore' | 'require'` giving even more control.

### Alternative

I guess there is a workaround but it's probably more complicated and for my use case I have yet to find one...

### Additional context

_No response_

### Validations

### Contributes

9romise:

- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ ] Ideally, include relevant tests that fail without this PR but pass with it.

### Description

This PR add support for `{ selector: string }` in `prev` and `next` options.

### Linked Issues

closes #1130, closes #1115, closes #1087, closes #1065, closes #956, closes #776, closes #442, closes #395, closes #311

### DIFF
--- a/packages/eslint-plugin/rules/padding-line-between-statements/padding-line-between-statements.ts
+++ b/packages/eslint-plugin/rules/padding-line-between-statements/padding-line-between-statements.ts
@@ -605,9 +605,9 @@ export default createRule<RuleOptions, MessageIds>({
           properties: {
             selector: {
               type: 'string',
-              required: true,
             },
           },
+          required: ['selector'],
           additionalProperties: false,
         },
         statementMatcher: {


Harpush:
@9romise Can this also apply to lines-between-class-members?

9romise:
> @9romise Can this also apply to lines-between-class-members?

I think it can. Would you like to have a try?

Harpush:
I can try once this PR is merged

> > @9romise Can this also apply to lines-between-class-members?
> 
> I think it can. Would you like to have a try?

Harpush:
@9romise A question regarding this feature.
For my example:
```ts
export function abc() {

}

export function test(a: number, b: string): void;
export function test(a: number, b: string): void;
export function test(a: number | string, b: string | number): void {
  return;
}

export function bcd() {
  
}
```
Should I use this rule config? Will it work with `block-like` and `export`?:
```
{ blankLine: 'always', prev: '*', next: 'block-like'},
{ blankLine: 'always', prev: 'block-like', next: '*' },
{
  blankLine: 'never',
  prev: { selector: 'ExportNamedDeclaration[declaration.type="TSDeclareFunction"]' },
  next: { selector: 'ExportNamedDeclaration[declaration.type="TSDeclareFunction"]' },
},
{
  blankLine: 'never',
  prev: { selector: 'ExportNamedDeclaration[declaration.type="TSDeclareFunction"]' },
  next: { selector: 'ExportNamedDeclaration[declaration.type="FunctionDeclaration"]' },
}
```
I am just afraid `block-like`  will be problematic to do with `selector`.
Maybe something like:
```
{
  blankLine: 'always',
  prev: '*',
  next: { selector: 'ExportNamedDeclaration > .declaration', match: 'block-like' },
}
```
So it will pad anything before `block-like` inside `ExportNamedDeclaration`.

9romise:
I'm not entirely sure, but feel free to give it a try via [pkg-pr-new](https://github.com/eslint-stylistic/eslint-stylistic/pull/1134#issuecomment-3888417131), it should let you preview the changes. I'll take another look once I'm back at my laptop.

Harpush:
> I'm not entirely sure, but feel free to give it a try via [pkg-pr-new](https://github.com/eslint-stylistic/eslint-stylistic/pull/1134#issuecomment-3888417131), it should let you preview the changes. I'll take another look once I'm back at my laptop.

It seems to work. What I don't understand is why `block-like` works with exports while `function-overload` doesn't work with exports (without this PR). Isn't it a bug?

Harpush:
@9romise So will you merge and release this PR?

Perdolique:
Will it work with single/multiline prefixes as well? (singleline-selector, multiline-selector)?

9romise:
> Will it work with single/multiline prefixes as well? (singleline-selector, multiline-selector)?

@Perdolique Not currently, but I think we could add an option like `isSingleLine` to support this. For example:
`{ selector: '', isSingleLine: true }`

PR #1207, opened 2026-05-06
fix(brace-style): restore opening-brace check for TS interface and enum #1193

### LINKED ISSUE #1193: `object-curly-newline`: enforce Allman-style brace placement on TypeScript interface and enum declarations
The `brace-style` rule does not enforce brace placement on TypeScript `interface` and `enum` declarations. Setting the rule to `allman` has no effect on these constructs.

Config:

```ts
'@stylistic/brace-style': ['error', 'allman']
```

Input :

```ts
interface Baz {
    foo: string;
    bar: number;
}

enum Qux {
    A,
    B
}
```

Expected:

```ts
interface Baz
{
    foo: string;
    bar: number;
}

enum Qux
{
    A,
    B
}
```

But no error is reported, the rule does not visit these nodes.

In the deprecated `@typescript-eslint` rule, it handled this correctly by visiting `TSInterfaceBody`, `TSModuleBlock`, and `TSEnumDeclaration`. When the rule was migrated to `@stylistic`, only `TSModuleBlock` was carried over.

Here's the [@typescript-eslint rule]( https://github.com/typescript-eslint/typescript-eslint/blob/6548240e6733381bf306d0b1c0112654c7ef4cd5/packages/eslint-plugin/src/rules/brace-style.ts) 

PR #823 intentionally removed `TSInterfaceBody` and `TSEnumDeclaration` from brace-style, treating them as "object-like" nodes that should be handled by `object-curly-newline` instead. However, `object-curly-newline` only controls newlines inside the braces, not whether the opening brace is placed on a new line. This means there is currently no rule that can enforce Allman-style brace placement on interface and enum declarations.

wreuel:
### Description
PR #823 removed the `TSInterfaceBody` and `TSEnumDeclaration` visitors on the rationale that interface/enum bodies are object-like and should be governed by `object-curly-newline`. That is true for newlines _inside_ the braces, but `object-curly-newline` does not control whether the opening `{` itself sits on its own line — leaving Allman style unenforceable on TS interfaces and enums.

### Linked Issues
resolves #1193

### Additional context

Reintroduce the visitors with a narrower scope:

- Extract the opening-curly placement logic into a new `validateOpeningCurly` helper.
- Add `TSInterfaceBody` and `TSEnumBody` visitors that call only that helper, so `nextLineOpen` / `sameLineOpen` are reported but the `blockSameLine` / `singleLineClose` checks remain owned by `object-curly-newline`.
- Restore the relevant valid/invalid test cases for `interface` and `enum`, scoped to opening-brace behaviour.
- Document the split in the rule README.

## Summary by CodeRabbit

* **Documentation**
  * Added a subsection clarifying brace-placement behavior for TypeScript interfaces and enums and recommending enabling a complementary rule for newline formatting inside braces.

* **Tests**
  * Expanded test coverage with valid and invalid cases for interface and enum declarations across all supported brace styles, including single-line exceptions and autofix expectations.

* **Refactor**
  * Improved brace-style validation to separate opening-brace placement checks for TypeScript interface/enum bodies from other curly-pair rules.


9romise (commented):
Personally, I understand that some people may have a need for this style. However, I believe that as the default behavior, we should not enable checks for them. We could add an `overrides` option. It would be better to clearly state this in the documentation for this rule and related rules.

PR #1225, opened 2026-06-10
fix(indent): align wrapped `?:` of chained conditional types

### LINKED ISSUE #1216: `indent`: wrapped `?:` of a `TSConditionalType` indents against the test-clause column instead of the parent indent
### Validations

### Describe the bug

For a wrapped `?:` inside a `TSConditionalType`, the `:` continuation is indented relative to the *visual column* of the test clause instead of the logical indent of the line containing `?`. When the conditional shares a line with siblings to its left, their widths leak into the continuation indent and add spurious nesting.

Minimal repro with `'@stylistic/indent': ['error', 4, { VariableDeclarator: 1, MemberExpression: 1 }]`:

```ts
export type Helper<T> =
    | (T extends string ? readonly T[]
        : T extends number ? readonly T[]
        : T);
```

The rule reports:

> Expected indentation of 12 spaces but found 8.

By logical level:

- level 0: the type-alias statement
- level 1 (+4): the `|` operand (continuation of `=`)
- level 2 (+4): the `?:` continuation
- → `:` should be at column 9 (8 spaces)

The rule arrives at 12 because the test clause `T extends string` happens to start at column 9 (it shares a line with `| (`), and the rule treats `(test-clause column) + indent` as the target. That conflates the test clause's visual position with the conditional's nesting depth. The parens are grouping and do not introduce a level. The `|` does not introduce a level for its operand. Only the `?:` introduces a level for its continuation.

The same bug reproduces without parens or unions — any time the conditional shares a line with constructs to its left whose width pushes the test clause's column past `parent-indent + indent`, the `:` indent inherits the extra. With a long object key on the same line as a ternary, the `:` indent grows with the key length, which clearly can't be right.

### Expected behavior

A wrapped `?:` continuation should indent to `indent-of-the-line-containing-? + one indent step`, independent of where the test clause starts on that line. For the repro above, that's `4 + 4 = 8` spaces. This matches Prettier and dprint, and matches how the rule already handles other continuations (e.g. binary-operator continuations are computed against the parent statement's indent, not against the left-operand column).

### Practical impact

TS projects using Prettier or dprint have to disable `@stylistic/indent` on `.ts` files, because unions of conditional types — a very common shape, fight the formatter.

### Related

- #676 — `indent-binary-ops` mistreating TS unions; commenters there also report circular fixes between `@stylistic/indent` and `@stylistic/indent-binary-ops` on conditional-type + union shapes.
- #1075 — recent refactor of `TSConditionalType` handling (behaviour-preserving, didn't touch the column logic).

### Reproduction

Inline code repro above; reproduces in any project with the listed `@stylistic/indent` config. Happy to provide a Stackblitz if needed.

### System Info

```shell
Reproduces against current `main` (`@stylistic/eslint-plugin@6.0.0-beta.4`) and on the latest 5.x.
```

### Contributes

- [x] I am willing to submit a PR to fix this issue
- [x] I am willing to submit a PR with failing tests

lo1tuma:
A `TSConditionalType` chained in the alternate of another conditional type and written compactly (`test ? consequent` on one line) was treated as a deeper nesting level, so each wrapped `:` indented one step further than the previous. The continuation indent ended up derived from the nested conditional's own first token instead of the chain it belongs to.

Such a chain is really a single continuation level, so its `?:` now align with the enclosing chain, matching how dprint formats unions of conditional types. The fully broken-out form (a line break after every `?`, as Prettier emits) keeps its existing per-level indentation, and JS `ConditionalExpression` is untouched.

Closes #1216

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved indentation checking for chained TypeScript conditional types.
  * Corrected alignment of conditional type branches and `?`/`:` operators in compact nested layouts.
  * Updated automatic fixes to apply the expected indentation for these cases.
  * Added coverage for valid and invalid formatting scenarios.


9romise (approved):
Sorry for the late reply. I've been busy and it took me a while to fully understand what this PR and the related issue are about. Thank you so much for your contribution!

PR #1242, opened 2026-07-20
fix(multiline-ternary): ignore nested conditional expressions when ignoreJSX option is set

danyadev:
### Description

Fixes false positives for nested conditional expressions with ignoreJSX, for example:

image


9romise (approved):
Thank you!

aleclarson:
can we get a release with this? thanks!

PR #1281, opened 2026-09-14
feat(max-statements-per-line): make the rule fixable

### LINKED ISSUE #1250: `max-statements-per-line`: make the rule fixable

`max-statements-per-line` reports but cannot fix. With `max: 1` — the strictest setting, and the one a "one statement per line" house style uses — every report has exactly one mechanical resolution: move the extra statements onto their own lines. There is no judgement in that step, but it is manual today.

The rule's `meta` carries neither `fixable` nor `hasSuggestions` (checked in `5.10.0` and on `main`), so neither `--fix` nor the editor lightbulb offers anything. It is already `type: 'layout'`, so a whitespace fixer would slot into the existing `--fix-type layout` story without any recategorisation.

This was proposed twice in ESLint core and never turned down on technical grounds:

- [eslint/eslint#9063](https://github.com/eslint/eslint/issues/9063) — closed for lack of team consensus.
- [eslint/eslint#17177](https://github.com/eslint/eslint/issues/17177) — closed as a duplicate with *"we currently aren't adding any new options to stylistic rules due to the policy change"*, i.e. the policy that moved these rules here.

### Suggested solution

For each reported line, insert a line break before every statement past `max`, indented with the leading whitespace of the line being split. The common shapes are all safe:

```js
foo(); bar();              // → break between the two
if (x) foo(); bar();       // → `if (x) foo();` / `bar();`
if (x) foo(); else bar();  // → break before `else`
do foo(); while (c);       // → break before `while`
```

Explicit semicolons are already required between statements sharing a line, so the break cannot introduce an ASI hazard.

Four things that want a decision rather than an assumption:

1. **Indentation.** Copying the split line's own leading whitespace is enough to produce valid, roughly-formatted output; anything finer is `@stylistic/indent`'s job on a later pass.
2. **Comments between statements.** `foo(); /* why */ bar();` — the comment has to land on one side of the break. Attaching it to the following statement matches how it reads.
3. **Single-line blocks.** In `if (c) { foo(); bar(); }` a break between the two statements is syntactically fine but leaves `{` and `}` hanging on the outer lines. Whether the fixer also breaks after `{` and before `}` is a separate formatting decision (arguably `@stylistic/brace-style`'s), and leaving those reports unfixed initially would be a defensible first cut.
4. **`ignoredNodes`.** The fixer has to use the same accounting as the check, so statements the user opted out of counting are not the ones that get moved.

### Alternative

A formatter (Prettier, dprint) reflows this as a side effect, but that means adopting a whole formatter to resolve one rule — not an option in a config that uses stylistic rules precisely to avoid one.

Alternatively `meta.hasSuggestions` instead of `meta.fixable`: it leaves `--fix` untouched and puts the change behind an explicit action in the editor. That is the conservative version of the same request, and would already remove the manual step where it is most annoying.

### Additional context

The two open issues on this rule ([#295](https://github.com/eslint-stylistic/eslint-stylistic/issues/295), [#1198](https://github.com/eslint-stylistic/eslint-stylistic/issues/1198)) are about false positives, not about fixing — as far as I can tell this has not been raised here before.

anandghegde:
### Description

Adds `fixable: 'whitespace'` to `max-statements-per-line`. The fix puts a line break before the reported statement (the first one over `max` on the line). The new line gets the same leading whitespace as the line being split. ESLint's fix passes then take care of any further statements on that line.

```js
foo(); bar(); baz();     // -> foo();\nbar();\nbaz();
if (a) foo(); bar();     // -> if (a) foo();\nbar();
foo(); /* c */ bar();    // -> foo();\n/* c */ bar();   (comment stays with the next statement)
if (a) {} bar()          // -> if (a) {}\nbar()
```

It's deliberately conservative. It only fixes when:

- the token before the statement is `;` or `}`, so the break always falls between two sibling statements. That leaves out `else foo();`, `case x: foo();` and the first statement after `{`.
- the statement is at the top level, or its enclosing block/case starts on an earlier line. Single-line blocks like `if (a) { b(); c(); }`, `() => { a(); b(); }` or `case 1: foo(); break;` are still reported but not fixed, so we don't end up with half-split braces (that's really `brace-style`'s territory).

In all other cases the rule reports the same way it does today. The fixer works from the statement the rule already reports, so `ignoredNodes` counts the same way for both.

Since `;` or `}` already separates the two statements, adding a newline before the second one can't change how the code parses. That holds for semicolon-less code too.

Tests: I added `output` to the existing invalid cases that now get fixed, plus `output: null` on a few that must stay unfixed. I also added new cases for comments, semicolon-less code, `do…while`, `export`, labelled statements, multi-line vs single-line `case`, arrow bodies and `ignoredNodes`. `rules.md` and `metadata.ts` are regenerated with `pnpm run update`.

Verified locally: `pnpm test` (133 files / 18454 tests pass), `pnpm run test:fixtures`, `pnpm run lint` and `pnpm run typecheck` all pass. 16 of the rule's tests fail with the rule change reverted.

### Linked Issues

resolves #1250

### Additional context

Because `all` only includes fixable rules, this also adds `max-statements-per-line` (default `max: 1`) to that config.

## Summary by CodeRabbit

- **New Features**
  - The `max-statements-per-line` lint rule can now automatically fix supported formatting violations by moving statements onto separate lines.
  - The rule documentation and metadata now identify it as fixable.

- **Bug Fixes**
  - Improved handling of comments, blocks, loops, conditionals, switch cases, exports, labels, arrow functions, and ignored nodes during autofix.


antfu (approved):

anandghegde:
Thanks for the review, @antfu. One thing left over: `test-eslint9` is the only red check, and it looks like a hang rather than a real failure.

In run 34819716264 the job printed the vitest banner and then nothing at all for 12 minutes:

```
08:28:08  > vitest -c vitest.config.unit.ts
08:28:09  RUN  v4.1.11  /home/runner/work/eslint-stylistic/eslint-stylistic/packages
08:40:25  ##[error]The operation was canceled.
```

No test output, no failure — it just never got going. The three sibling `test` matrices in the same run finished green in ~2m30s, and `main` is green.

Could someone re-run that job? I'd rather not push a no-op commit and dismiss your approval to retrigger it.

